Issue #14, opened 2013-11-01
Rename readonly/writeonly to readable/writeable


lanthaler:
We currently have two properties in the Hydra vocabulary which I'm not entirely happy with yet: `hydra:readonly` and `hydra:writeonly`. I think more and more that renaming them to `hydra:readable` and `hydra:writeable` would make them easier to work with.

The reason why I gave them this name in the first place was because I thought that it is only necessary to explicitly set them if a property is either readonly or writeonly but that's not entirely true due to the open world assumption Hydra is based on. If it's not set, you can't assume it is set to false. You have to accept that you don't know what value it has.

The advantage of renaming them would be that the inconsistency which is cause if something is marked as both read- and writeonly would be eliminated and programming against them would become easier. Apart from introducing a breaking change (which is still quite cheap at this point in time) I don't see any downside of the change but maybe I miss something.

PR #83, opened 2015-01-05
Change readonly/writeonly to readable/writeable.

RubenVerborgh:
Implements the proposed resolution for #14.

Note that I have explicitly mentioned “readable/writeable **for a client**”, as we discussed that these properties may depend on the client accessing the response (like hypermedia controls do).

### DIFF
--- a/spec/latest/core/core.jsonld
+++ b/spec/latest/core/core.jsonld
@@ -19,8 +19,8 @@
     "SupportedProperty": "hydra:SupportedProperty",
     "property": { "@id": "hydra:property", "@type": "@vocab" },
     "required": "hydra:required",
-    "readonly": "hydra:readonly",
-    "writeonly": "hydra:writeonly",
+    "readable": "hydra:readable",
+    "writeable": "hydra:writeable",
     "supportedOperation": { "@id": "hydra:supportedOperation", "@type": "@id" },
     "Operation": "hydra:Operation",
     "CreateResourceOperation": "hydra:CreateResourceOperation",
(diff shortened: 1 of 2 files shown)


spec/latest/core/index.html
```diff
@@ -490,8 +490,8 @@
             ****"@type": "SupportedProperty"****,
             ****"property"****: "#property", ####// The property####
             ****"required"****: true, ####// Is the property required in a request to be valid?####
-            ****"readonly"****: false, ####// Can the property's value be modified or is it read-only?####
-            ****"writeonly"****: true ####// Can the property's value be retrieved or is it write-only?####
+            ****"readable"****: false, ####// Can the client read the property's value?####
+            ****"writeable"****: true ####// Can the client write the property's value?####
```
lanthaler: What about changing the comments to "Can *a* client *retrieve/change* the property's value?"

spec/latest/core/core.jsonld
```diff
@@ -180,19 +180,19 @@
       "vs:term_status": "testing"
     },
     {
-      "@id": "hydra:readonly",
+      "@id": "hydra:readable",
       "@type": "rdf:Property",
       "label": "ready-only",
```
lanthaler: Please also update the label; same below

spec/latest/core/core.jsonld
```diff
@@ -180,19 +180,19 @@
       "vs:term_status": "testing"
     },
     {
-      "@id": "hydra:readonly",
+      "@id": "hydra:readable",
       "@type": "rdf:Property",
       "label": "ready-only",
-      "comment": "True if the property is read-only, false otherwise.",
+      "comment": "True if the client can read the property, false otherwise.",
```
lanthaler: Same as in index.html - but I read this afterwards :-)

"True if *a* client can *retrieve/change* the property's value, false otherwise".

spec/latest/schema.org/index.html
```diff
@@ -446,16 +446,16 @@
           <td>hydra:required</td>
         </tr>
         <tr>
-          <td>readonly</td>
+          <td>readable</td>
```
lanthaler: As the note at the beginning of this documents states, this is document "serves purely as a historical reference". As such, we don't update it anymore. Please rever the changes to this file.

spec/latest/core/core.jsonld
RubenVerborgh: Done; amended as f83cb47042a7cd724b650245f5287589826d8ad7.

spec/latest/core/core.jsonld
RubenVerborgh: Done; amended as f83cb47042a7cd724b650245f5287589826d8ad7.

spec/latest/core/index.html
RubenVerborgh: Done; amended as f83cb47042a7cd724b650245f5287589826d8ad7.

spec/latest/schema.org/index.html
RubenVerborgh: My `grep` missed this :smile:

Reverted; amended as f83cb47042a7cd724b650245f5287589826d8ad7.

lanthaler:
Thanks a lot Ruben!

PR #84, opened 2015-01-05
Rename statusCodes and StatusCodeDescription.

### LINKED ISSUE #27: Having both hydra:statusCode and hydra:statusCodes should be avoided
***[Raised by Ruben Verborgh](http://lists.w3.org/Archives/Public/public-hydra/2014Jan/0045.html) as part of issue #25:***

The case of `statusCodes` shows to how much confusion this leads:
we have both a `statusCode` and a `statusCodes` property.

The situation would be that a resource can lead to `statusCodes` X, Y, Z, each of which have a `statusCode` 200, 400, 500, respectively.

I think the correct way is that `statusCodes` becomes `status`, and that the current `statusCode` remains the way it is.

---
An alternative would be to rename it to `possibleStatus` or `additionalStatusInformation`.

RubenVerborgh:
Implements the proposed resolution for #27.

The last comment in #27

> Should hydra:returns be moved into a Status (see http://lists.w3.org/Archives/Public/public-hydra/2014Sep/0089.html)

has not reached consensus yet; I propose to close #27 and move the comment to a new issue.

### DIFF
--- a/spec/latest/core/core.jsonld
+++ b/spec/latest/core/core.jsonld
@@ -29,8 +29,8 @@
     "method": "hydra:method",
     "expects": { "@id": "hydra:expects", "@type": "@vocab" },
     "returns": { "@id": "hydra:returns", "@type": "@vocab" },
-    "statusCodes": { "@id": "hydra:statusCodes", "@type": "@id" },
-    "StatusCodeDescription": "hydra:StatusCodeDescription",
+    "possibleStatus": { "@id": "hydra:possibleStatus", "@type": "@id" },
+    "Status": "hydra:Status",
     "statusCode": "hydra:statusCode",
     "Error": "hydra:Error",
     "Resource": "hydra:Resource",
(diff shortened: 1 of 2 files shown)


spec/latest/core/core.jsonld
```diff
@@ -139,11 +139,11 @@
       "vs:term_status": "testing"
     },
     {
-      "@id": "hydra:statusCodes",
+      "@id": "hydra:possibleStatus",
       "@type": "hydra:Link",
       "label": "status codes",
       "comment": "Additional information about status codes that might be returned by the Web API",
```
RubenVerborgh: Done and amended as 0b3e12fe2a7609fb5094888a028a495b637a3e3c.

I did dropped ”description of” though (use / mentions) and made it singular, in accordance with other properties:

> A status that might be returned by the Web API (other statuses should be expected and properly handled as well)

spec/latest/core/index.html
```diff
@@ -432,7 +432,7 @@
         ****"supportedClass"****: [
           ####... Classes known to be supported by the Web API ...####
         ],
-        ****"statusCodes"****: [
+        ****"possibleStatus"****: [
           ####... Additional information about HTTP status codes ... ####
```
RubenVerborgh: Done; amended as 0b3e12fe2a7609fb5094888a028a495b637a3e3c.

spec/latest/core/index.html
```diff
@@ -545,7 +545,7 @@
             "method": "POST",
             "expects": "http://api.example.com/doc/#Comment",
             "returns": "http://api.example.com/doc/#Comment",
-            "statusCodes": [****
+            "possibleStatus": [****
               ####... optional information about status codes that might be returned ...####
```
RubenVerborgh: Done; amended as 0b3e12fe2a7609fb5094888a028a495b637a3e3c.

spec/latest/core/core.jsonld
lanthaler: OK, that’s fine as well.

Issue #80, opened 2015-01-04
Introduce VariableRepresentation class


lanthaler:
In #30 we introduced `hydra:BasicRepresentation` and `hydra:ExplicitRepresentation` but didn't describe them any further for machines. For certain use cases (mostly searching available representation formats) it would be beneficial to have a class such as `hydra:VariableRepresentation`.

lanthaler:
Generally, we try to avoid having a property and a class which differ only in their capitalization. Should we follow Schema.org's convention and name this `hydra:VariableRepresentationType` instead of just `hydra:VariableRepresentation`?

lanthaler:
Call for consensus: http://lists.w3.org/Archives/Public/public-hydra/2015Jan/0204.html

lanthaler:
RESOLVED: Add the class `VariableRepresentation`. The only two instances at the moment are `hydra:BasicRepresentation` and `hydra:ExplicitRepresentation`.

PR #85, opened 2015-01-05
Introduce VariableRepresentation class

RubenVerborgh:
Implements the proposed resolution for #80.

### DIFF
--- a/spec/latest/core/core.jsonld
+++ b/spec/latest/core/core.jsonld
@@ -457,7 +457,6 @@
     {
       "@id": "hydra:BasicRepresentation",
       "@type": "hydra:VariableRepresentation",
-      "subClassOf": "hydra:Resource",
       "label": "basic value representation",
       "comment": "A representation that serializes just the lexical form, but omits language and type information.",
       "vs:term_status": "testing"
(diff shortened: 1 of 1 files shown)


spec/latest/core/core.jsonld
```diff
@@ -442,19 +443,28 @@
...
       "@type": "hydra:Class",
       "subClassOf": "hydra:Resource",
+      "label": "representation of variable values",
+      "comment": "A representation specifies how to serialize variable values into strings.",
+      "vs:term_status": "testing"
+    },
+    {
+      "@id": "hydra:BasicRepresentation",
+      "@type": "hydra:VariableRepresentation",
+      "subClassOf": "hydra:Resource",
```
lanthaler: We is this a sub class of `hydra:Resource` and not an instance thereof? Same below

RubenVerborgh: Those lines were still remains of the previous version; 9e267985fdc0ef7e84c5d19a0aae4bdeb5f565e6 removes them. It's probably not necessary to explicitly list `hydra:Resource` as type, since we now have `hydra:VariableRepresentation` as the more specific type.

Issue #11, opened 2013-10-02
Remove predefined operations from the core vocabulary.


lanthaler:
There have been some discussions about the [use](http://lists.w3.org/Archives/Public/public-hydra/2013Oct/0001.html) and [consistency](http://lists.w3.org/Archives/Public/public-hydra/2013Sep/0011.html) of the pre-defined operations (`CreateResourceOperation`, `ReplaceResourceOperation`, and `DeleteResourceOperation`).

The only reason they have been defined was to help developers to get started. This may not be a strong enough argument to keep the in the core vocabulary.

Issue #2, opened 2013-07-15
Use the term Action instead of Operation


lanthaler:
> Hydra has "operations" whereas
> Siren has "actions". Activity Streams also has actions. So has the latest
> schema.org proposal. I'm not a native speaker and don't really care how this
> is called but I realized recently that almost everyone else calls these
> things "actions" :-)

[http://lists.w3.org/Archives/Public/public-hydra/2013Jul/0002.html]

tpluscode:
This has been here for over 3 years. Is it fair to say that the term *Operation* has already gained a foothold in Hydra's spec? I think it's not worth changing it now any more.

akuckartz:
No such terms in Hydra should be considered final.

lanthaler:
Yeah, IMO we should keep operation to describe HTTP requests etc. and may use he term action for the abstract act (like something, order something, …)

tpluscode:
Sounds good. But would the `action` become a term used in the vocabulary?

RubenVerborgh:
FYI I have called them “actions” in the [draft architectural diagram](https://github.com/RubenVerborgh/Hydra-Architecture-Diagram/) because this sounded most natural in that context. (Of course this doesn't represent a decision in any way.)

elf-pavlik:
Short note on Operations (on web resources) and Actions (on abstract things). I've mostly seen instances of `hydra:Operation` represented with blank nodes but they could as well have `@id`. Action on abstract thing could possibly reference particular Operation (since each operation could have an `@id`) and web resource on which to perform it - (maybe with https://schema.org/target )

Taking example of Alice and list of her 'followers', represented by https://www.w3.org/community/hydra/wiki/Collection_Design and used like [ldp:IndirectContainer](https://www.w3.org/TR/ldp/#ldpic)

```json
{
  "@context": { ... },
  "@graph": [
    {
      "@id": "https://dataset.example/api#operation-123",
      "@type": "hydra:CreateResourceOperation",
      "hydra:method": "POST",
      "hydra:expects": "foaf:PersonalProfileDocument"
    },
    {
      "@id": "https://dataset.example/alice",
      "@type": [ "hydra:Resource", "foaf:PersonalProfileDocument" ],
      "foaf:primaryTopic": {
        "@id": "https://idp.example/alice",
        "foaf:name": "Alice",
        "schema:potentialAction": [
          {
            "@type": "schema:FollowAction",
            "schema:target": "https://dataset.example/alice/followers/",
[...]

PR #86, opened 2015-01-05
Implement “remove predefined operations”

RubenVerborgh:
This pull request **partially** addresses #11 by removing the predefined operations from the vocabulary.
However, I need some input to finalize this.
1. Should `hydra:Operation` be removed/replaced? And if so, what about `hydra:supportedOperation`, `hydra:operation`, `hydra:method`, `hydra:expects`, `hydra:returns`?
2. In the specification document, should we remove explanations about `CreateResourceOperation`, `ReplaceResourceOperation` and `DeleteResourceOperation`? Or should we, as suggested in #11, replace them by Schema.org explanations? The latter, however, need not necessarily be part of the spec.

As a temporary resolution for the second point, I propose to remove. We can always add explanations later (but perhaps already create an issue for that then).


luchsamapparat:
Hi,

is there any roadmap on when to expect this Change? I'm currently exploring Hydra and I'm unsure how to proceed

tpluscode:
@RubenVerborgh Why do you propose removing `hydra:Operation` and others? Following [this comment ](https://github.com/HydraCG/Specifications/issues/2#issuecomment-269014538) in #2 I think that operations do have their place describing HTTP protocol and actions being used in API-specific terms.

For second point, I think it may be good to actually add schema actions as illustrations of suggested usage.

RubenVerborgh:
@tpluscode I think the question should be the other way round: why do we need them? As announced on the mailing list, I'd like to take a step back and start with a clean slate from the lessons we learned. It might well be that we need something like operation.

(Sketching a proposed architectural diagram as we speak, I expect to send this out today.)

RubenVerborgh:
[Architectural diagram (draft)](https://github.com/RubenVerborgh/Hydra-Architecture-Diagram/)

lanthaler:
I'm fine with removing CreateResourceOperation, ReplaceResourceOperation, and DeleteResourceOperation. I agree with @tpluscode that we will likely need `hydra:Operation` etc. Should we proceed with merging this PR and discuss the rest separately? 

Unless I hear objections, I'll merge this PR on the weekend (we previously agreed to delete them in #11).

asbjornu:
👍  for deleting the concrete operations while keeping the generic `hydra:Operation`.

Issue #17, opened 2013-11-22
Introduce a datatype for RFC6570 IRI templates


lanthaler:
***This was proposed by @rubenverborgh [on the mailing list](http://lists.w3.org/Archives/Public/public-hydra/2013Nov/0037.html):***

> No, I'm not aware of any type representing RFC6570 IRI templates and didn't
> see the need to complicate Hydra by defining one :-)

It's not as much about complicating as about being exact.
The thing is not a string, it's an RFC6570 template.
Please seize this opportunity :-)

Note that this even enables future extensibility automatically.
If RFC25000 suddenly comes up with another template,
it can be supported out of the box:

```
<x> hydra:template "old-template"^^hydra:rfc6570Template.
<x> hydra:template "new-template"^^rfc25000:template.
```

So the same resources can have multiple templates in different standards.
That's extensibility.

lanthaler:
Results of the straw poll regarding the capitalization of datatypes: http://lists.w3.org/Archives/Public/public-hydra/2015Jan/0156.html

lanthaler:
Call for consensus: http://lists.w3.org/Archives/Public/public-hydra/2015Jan/0157.html

lanthaler:
RESOLVED: The name of the datatype for RFC6570 IRI templates is `hydra:Rfc6570Template`. Other datatypes will follow this naming pattern.

PR #87, opened 2015-01-05
Add hydra:Rfc6570Template.

RubenVerborgh:
Implements the proposed resolution for #17.

### DIFF
--- a/spec/latest/core/core.jsonld
+++ b/spec/latest/core/core.jsonld
@@ -50,6 +50,7 @@
     "TemplatedLink": "hydra:TemplatedLink",
     "IriTemplate": "hydra:IriTemplate",
     "template": "hydra:template",
+    "Rfc6570Template": "hydra:Rfc6570Template",
     "variableRepresentation": { "@id": "hydra:variableRepresentation", "@type": "@vocab" },
     "BasicRepresentation": "hydra:BasicRepresentation",
     "ExplicitRepresentation": "hydra:ExplicitRepresentation",
@@ -431,9 +432,16 @@
       "@id": "hydra:template",
       "@type": "rdf:Property",
       "label": "template",
+      "comment": "A templated string with placeholders. The literal's datatype indicates the template syntax; if not specified, hydra:Rfc6570Template is assumed.",
+      "seeAlso": "hydra:Rfc6570Template",
+      "vs:term_status": "testing"
+    },
+    {
+      "@id": "hydra:Rfc6570Template",
+      "@type": "rdfs:Datatype",
+      "label": "RFC6570 IRI template",
       "comment": "An IRI template as defined by RFC6570.",
       "seeAlso": "http://tools.ietf.org/html/rfc6570",
-      "domain": "hydra:IriTemplate",
       "range": "xsd:string",
       "vs:term_status": "testing"
     },
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -723,11 +723,14 @@
       is to give the client a template to construct such a URL at runtime.
       In Hydra, the <i>IriTemplate</i> class is used to do so.</p>
 
-    <p>A <i>IriTemplate</i> consists of a <i>template</i>, whose syntax is
-      specified in [[!RFC6570]], and a set of <i>mappings</i>. Each
-      <i>IriTemplateMapping</i> maps a <i>variable</i> used in the template
+    <p>An <i>IriTemplate</i> consists of a <i>template</i> literal
+      and a set of <i>mappings</i>.
+      Each <i>IriTemplateMapping</i> maps a <i>variable</i> used in the template
       to a <i>property</i> and may optionally specify whether that variable
-      is <i>required</i> or not.</p>
+      is <i>required</i> or not.
+      The syntax of the template literal is specified by its datatype
+      and defaults to the [[!RFC6570]] URI Template syntax,
+      which can be explicitly indicated by <i>hydra:Rfc6570Template</i>.</p>
 
     <pre class="example" data-transform="updateExample"
          title="Description of an IRI Template">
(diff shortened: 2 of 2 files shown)


RubenVerborgh:
I'd be included to keep it as-is; or either `RFC` in all-caps. Half-caps don't work for an acronym IMHO.

spec/latest/core/core.jsonld
```diff
@@ -431,6 +432,16 @@
       "@id": "hydra:template",
       "@type": "rdf:Property",
       "label": "template",
+      "comment": "A templated string with placeholders. The literal's datatype indicates the template syntax; if not specified, hydra:rfc6570Template is assumed.",
+      "seeAlso": "hydra:rfc6570Template",
+      "domain": "hydra:IriTemplate",
+      "range": "rdfs:Literal",
```
RubenVerborgh: Agree on dropping the domain.
I strongly recommend to keep the range. All this says is that it's a literal property, and I know only one case where it makes sense to allow both objects and literals as a property's value (`rdf:object`).

Amended as 6df5dfc49b677ccc21d68e2a1c5c725954a532a1.

spec/latest/core/core.jsonld
```diff
@@ -431,6 +432,16 @@
...
+      "comment": "A templated string with placeholders. The literal's datatype indicates the template syntax; if not specified, hydra:rfc6570Template is assumed.",
+      "seeAlso": "hydra:rfc6570Template",
+      "domain": "hydra:IriTemplate",
+      "range": "rdfs:Literal",
+      "vs:term_status": "testing"
+    },
+    {
+      "@id": "hydra:rfc6570Template",
+      "@type": "rdfs:Datatype",
+      "label": "template",
```
RubenVerborgh: Done; amended as 6df5dfc49b677ccc21d68e2a1c5c725954a532a1.

spec/latest/core/core.jsonld
lanthaler: Really? Dropping the range would allow us to support templates which aren’t expressed in the form of a string (which needs to be parsed) but as a “structured value” which might be easier to process in a client.

akuckartz:
My preference: "RFC6570Template".

But I suggest to discuss capitalization conventions in the broader Semantic Web and Linked Data communities.

spec/latest/core/core.jsonld
RubenVerborgh: Mwah, I think that's already the job of `hydra:IriTemplate` itself. The `hydra:template` property really points to a skeleton. Syntaxes can differ, but they will be strings.

lanthaler: Not necessarily I’d say. What can *machines* do in practice if we add this range that they can't otherwise?

RubenVerborgh: Assume it's a literal property, and thus perform literal operations. Although it's probably mostly handy for humans writing program code against this property.

While I do think properties should at least be either literals or objects, I'm not fighting for this one. Whatever resolution you prefer is fine with me.

lanthaler: I would like to keep as many doors open as possible to allow usages we haven't anticipated. We can refine this later if necessary. So I'd suggest to drop the range for the time being.

RubenVerborgh: Alright; amended to 71224463766f3b087aa1fc49db2e947625d62519.

lanthaler: Great! Thanks. So the only remaining question is the capitalization. I’ll send out a straw poll to move this forward.

spec/latest/core/core.jsonld
```diff
@@ -50,6 +50,7 @@
     "TemplatedLink": "hydra:TemplatedLink",
     "IriTemplate": "hydra:IriTemplate",
     "template": "hydra:template",
+    "rfc6570Template": "hydra:rfc6570Template",
```
lanthaler: Please update the capitalization

spec/latest/core/index.html
```diff
@@ -723,11 +723,14 @@
       is to give the client a template to construct such a URL at runtime.
       In Hydra, the <i>IriTemplate</i> class is used to do so.</p>
 
-    <p>A <i>IriTemplate</i> consists of a <i>template</i>, whose syntax is
-      specified in [[!RFC6570]], and a set of <i>mappings</i>. Each
-      <i>IriTemplateMapping</i> maps a <i>variable</i> used in the template
+    <p>An <i>IriTemplate</i> consists of a <i>template</i> literal
+      and a set of <i>mappings</i>.
```
lanthaler: ... and remove those linebreaks if...

spec/latest/core/index.html
```diff
@@ -723,11 +723,14 @@
...
+      Each <i>IriTemplateMapping</i> maps a <i>variable</i> used in the template
       to a <i>property</i> and may optionally specify whether that variable
-      is <i>required</i> or not.</p>
+      is <i>required</i> or not.
+      The syntax of the template literal is specified by its datatype
```
lanthaler: ... they aren't the end of a paragraph. Makes it much easier to scan when reading a source. Thanks!

spec/latest/core/core.jsonld
RubenVerborgh: Amended in 690c4f2c7bd619d7dd50ff0eb78fd541544a110d.

spec/latest/core/index.html
RubenVerborgh: The line gets too long if I do; similar line wrapping is applied on the rest of the file.

spec/latest/core/index.html
RubenVerborgh: The line gets too long if I do; similar line wrapping is applied on the rest of the file.

lanthaler:
Thanks Ruben

PR #97, opened 2015-02-02
Explicitly mention namespace in Core Vocabulary document

RubenVerborgh:
The current Core Vocabulary document did not explicitly mention the namespace and suggested prefix.

### DIFF
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -169,6 +169,10 @@
     explicitly marked as non-normative. Everything else in this
     specification is normative.</p>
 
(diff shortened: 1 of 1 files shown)


lanthaler:
Good idea, but IMO the section "Hydra at a Glance" would be more appropriate than the "Conformance" section. What do you think?

elf-pavlik:
Maybe we could merge this one just not to have stale open PRs and once we pick up editorial work again we can always move things between sections.

lanthaler:
I'll merge it and move it to the other section.

PR #103, opened 2015-11-09
Lexicographica

jaw111:
Remove some gerunds

### DIFF
--- a/spec/latest/triple-pattern-fragments/index.html
+++ b/spec/latest/triple-pattern-fragments/index.html
@@ -82,8 +82,8 @@
         And what does “knowing” mean in this context?
       </p>
       <p>
-        By giving each of those three pieces of information a URL,
-        we can identify them clearly:
+        We use URLs to clearly and unambiguosly identify each of those
+        three pieces of information:
       </p>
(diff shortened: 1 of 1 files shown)


RubenVerborgh:
Thanks a lot!

spec/latest/triple-pattern-fragments/index.html
```diff
@@ -229,7 +229,7 @@
         <dt>hypermedia controls</dt>
         <dd>
           A Triple Pattern Fragment MUST contain hypermedia controls
-          that allow retrieving any other Triple Pattern Fragment of the same dataset.
+          that allow retrieval of any other Triple Pattern Fragment of the same dataset.
```
vanthome: Any reasoning why the continuous tense is wrong here? I learned that it makes documents more lively and is thus to be preferred.

jaw111: Depends on writing style. I had learned from Simplified Technical English that use of gerunds is to be avoided as they can cause confusion for non-native readers (and can be awkward to translate). Downside is it leads to a 'drier' style, but if that aids comprehension then it's fine with me.

vanthome: ok, thanks for the explanation!

PR #106, opened 2015-12-10
improved examples using Comment class

elf-pavlik:
During presentation at http://paris.apidays.io/ last Tuesday I showed examples of defining properties of Comment class and operation on collection of comments.
http://www.slideshare.net/elfpavlik/api-standardization-work-in-w3c-groups/94

IMO with modifications proposed in his PR one could explain it much more easily. Now it also encourages to reuse existing vocabularies like schema.org. Which matches @lanthaler's presentation from last year http://www.slideshare.net/lanthaler/the-web-is-changing-from-strings-to-things/8

### DIFF
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -476,11 +476,11 @@
       associated with.</p>
 
     <pre class="example" data-transform="updateExample"
-         title="Defining a class and documenting its supported properties">
+         title="Defining/augmenting a class and documenting its supported properties">
       <!--
       {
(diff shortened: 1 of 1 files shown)


lanthaler:
I imagine this will result in many comments regarding "poluting" the vocabulary definition etc. and that the class should be subclassed instead. Let's see if we come up with a more elegant design when we discuss operations (or are able to reuse SHACL or something) instead of doing of quick fix now. WDYT?

elf-pavlik:
SGTM, maybe worth adding note about planned changes inline in the spec?

PR #110, opened 2017-01-10
Proof of concept for authoring spec in markdown

tpluscode:
In the previously used version of ReSpec only inline markdown was supported. With great help from @marcoscaceres, now it's possible to create specification sections in imported markdown files (see w3c/respec#993)

Please let me know if you'd like to explore splitting the spec in smaller chunks.

Also, I refactored the `respec-w3c-extensions.js` file to be independent of jQuery (see [this comment](https://github.com/w3c/respec/issues/993#issuecomment-269773682))

### DIFF
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -74,7 +74,9 @@
           // { uri: "diff-20120712.html", label: "diff to previous version" }
           { uri: "core.jsonld", label: "JSON-LD" },
           { uri: "core.ttl", label: "Turtle" },
-      ]
+      ],
+    
+      format: "markdown"
   };
 //]]>
 </script>
@@ -950,12 +952,12 @@
 </section>
 
 <section class="informative">
-  <h2>Acknowledgements</h2>
+  ## Acknowledgements
 
-  <p>The authors would like to thank the following individuals for contributing
-    their ideas and providing feedback for writing this specification:
-    Arnau Siches, Mark Baker, Martijn Faassen, Matthias Lehmann, Ruben Verborgh,
-    Ryan J. McDonough, Sam Goto, Thomas Hoppe, @wasabiwimp (on GitHub).</p>
+  The authors would like to thank the following individuals for contributing
+  their ideas and providing feedback for writing this specification:
+  Arnau Siches, Mark Baker, Martijn Faassen, Matthias Lehmann, Ruben Verborgh,
+  Ryan J. McDonough, Sam Goto, Thomas Hoppe, @wasabiwimp (on GitHub).
 </section>
 
 <section class="appendix informative">
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -166,26 +166,7 @@
     granularity.</p>
 </section>
 
-<section>
-  <h2>Conformance</h2>
-
-  <p>This specification describes the conformance criteria for
-    <dfn data-lt="Hydra API documentation">Hydra API documentations</dfn>
-    and <dfn data-lt="Hydra client">Hydra clients</dfn>. These criteria are
-    relevant to authors, authoring tool implementers, and client
-    implementers. All authoring guidelines, diagrams, examples, and notes
-    in this specification are non-normative, as are all sections
-    explicitly marked as non-normative. Everything else in this
-    specification is normative.</p>
-
-  <p class="issue">Conformance for Hydra clients should probably not be
-    specified in this document.</p>
-
-  <p class="issue">Add normative statements</p>
-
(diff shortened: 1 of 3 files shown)


RubenVerborgh:
While this is great, we should probably differentiate between two goals:
- drafting the elements of Hydra in Markdown (without bothering about language etc. as it will appear in the spec, just getting the ideas out)
- writing the spec

So I see this work in the second context; is this what you have in mind?

tpluscode:
Yes, I had the actual spec in mind. "Bothering about language" doesn't interfere with ease of authoring. Markdown and/or splitting the HTML in smaller chunks could make it easier to author the specs.

spec/latest/core/index.html
```diff
@@ -3,6 +3,7 @@
 <head>
 <title>Hydra Core Vocabulary</title>
 <meta charset="utf-8">
+<script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=fetch|gated"></script>
```
lanthaler: How is this better if it requires an additional dependency (at least in some browsers) and makes more HTTP requests (in all browsers)?

spec/latest/core/index.html
```diff
@@ -164,26 +167,7 @@
...
-
-  <p class="issue">Conformance for Hydra clients should probably not be
-    specified in this document.</p>
-
-  <p class="issue">Add normative statements</p>
-
-  <p>The key words MUST, MUST NOT, REQUIRED, SHALL, SHALL NOT, SHOULD,
-    SHOULD NOT, RECOMMENDED, NOT RECOMMENDED, MAY, and OPTIONAL in this
-    specification have the meaning defined in [[!RFC2119]].</p>
+<section id="conformance" data-include="section/conformance.html">
```
lanthaler: Not a big fan of these includes. They work pretty badly when working offline. You are forced to bring up a server to get them working

lanthaler:
I’m not sure what the desired goal of this proof of concept is. Do you want to convert the entire spec to Markdown? Do you want to split it into multiple files? Both?

spec/latest/core/index.html
tpluscode: Re: the dependency - you could use pure XHR instead of fetch. Just requires more code but is doable. The polyfill would be required anyway, because ReSpec internally requires Promises. On the other hand ReSpec doesn't support IE anyways ATM so we could just drop that anyway actually

Re: more requests (speaking about imports not poylfill) - not sure if it's a big deal. They will all be small and nicely cacheable. 

And doesn't ReSpec have a `build` facitily to combine all those files together in a ready to publish form?

spec/latest/core/index.html
tpluscode: Forgive me saying, but this is just whining :). How problematic is it to use [local-web-server](https://www.npmjs.com/package/local-web-server) or [http-server](https://www.npmjs.com/package/http-server)?

lanthaler: Maybe but I don't see any value in splitting a HTML file which isn't even a 1000 lines long into multiple files which then require a server running to be able to work with it.

tpluscode:
Probably both with the preference of splitting if first into multiple files. Smaller files could be easier to manage than one large HTML document, with all that indentation and pointy markup. And markdown is just easier to edit. 

And contribute. That is the real goal. Lower the barrier.

spec/latest/core/index.html
lanthaler: Sure, there are multiple ways. My argument however was that adding this dependency doesn't result in benefits that make it worthwhile.

asbjornu:
I approve of this pull request, or at least of the direction it is moving. 😄  👍  I too would like the spec to be entirely written in Markdown and while the spec currently is less than 1000 LOC, it needs to grow to incorporate informative examples and such. In that future scenario, managing a larger set of smaller, feature-focused files, is easier than one gigantic one.

spec/latest/core/index.html
asbjornu: Being able to use `Fetch` and `Promise` makes it worthwhile, I think. Or should we require a certain browser version to be able to read the specification?

tpluscode: ReSpec requires Promises, which makes it incompatible with IE anyway. And I wouldn't fight over fetch.

lanthaler: ReSpec was just updated to accept Promises in the pre- and postprocess hooks. See w3c/respec#1034. Could you please file a separate PR which does that preprocess step (I changed it to postprocess on the weekend before ReSpec was updated). Bonus points for IE support.
We can then discuss the splitting and Markdown conversion separately.
Thanks

PR #111, opened 2017-03-25
[client-interface] Suggestion for browser-like client API

dschulten:
PR to show how a browser-like client could behave


RubenVerborgh:
For what languages is the interface?

Is an entrypoint necessary?

What is the "current" representation?

dschulten:
The sample cli is written in pseudocode. It makes use of a hypothetical hydra object which has certain property accessors and methods. Notably:

- setLocation: the equivalent of entering a url in the address bar
- representation: access to the header and body received after requesting the most recent resource, i.e. the current application state
- controls: access to hypermedia affordances in the current representation
- execute: make use of a given hypermedia affordance to transition to another representation, i.e. application state.

The interaction with the hypothetical hydra client uses these accessors and methods by means of the pseudo language. A real hydra object would of course need accessors and methods for a real language.

The findEntrypoint method is not necessary, as stated by the hydra spec.

The current representation is usually the result of an interaction with a server resource identified by a URI (unless the interaction was fulfilled from a local cache). E.g. a GET for a URI may result in a response which is a representation of the requested resource.

drafts/client-interface.md
```diff
@@ -3,11 +3,150 @@
...
+```
+hydra.setLocation(apiurl);
+// apiurl is now the current representation
+
+// we want to retrieve or download the menu of the food establishment
+menu = hydra.controls["http://schema.org/hasMenu"]
+if(menu && menu.methods["GET"])
+    hydra.execute("GET", menu.href)
+    print hydra.representation
+    hydra.back()
```
lanthaler: I'd rather store state explicitly in and keep hydra stateless as already mentioned above. The also reduces the cognitive overhead of keeping in mind what has been done (tricky when there are multiple branches or error handling)

drafts/client-interface.md
```diff
@@ -3,11 +3,150 @@
...
+Assuming that apiurl is known as API entrypoint and that it is a [FoodEstablishment](http://schema.org/FoodEstablishment)
 
 Related action: [DownloadAction](http://schema.org/DownloadAction)
 
+The client has a downloadAction() method for downloading a resource to a file rather than making that 
+resource the new current resource. The client looks for the operation of type DownloadAction 
+on the given control and executes its method.
+
+This assumes that the Hydra client shall have a limited number of convenience methods which make use of 
+known schema.org Actions.
```
lanthaler: I'd argue that for this document we should keep that out of scope and should instead use something like your `.controls["http://schema.org/hasMenu"]` construct. I'd rather allow to abbreviate that by specifying a context for instance. Separate methods can be created by wrapping/subclassing these interfaces

drafts/client-interface.md
```diff
@@ -3,11 +3,150 @@
...
+            print "enter " + expectedProperty.property + 
+                "[default " + expectedProperty.defaultValue + "]" +
+                expectedProperty.required ? "(*)" : ""
+            input value
+            data[expectation.kind][expectation.variable] = value
+    hydra.execute("POST", uri, headers={}, data["body"])
+    print hydra.representation
+```
+
+## Working with Resources in a URI space
```
lanthaler: "Working with URI Templates" (or IRI Templates) would likely be much easier to understand

drafts/client-interface.md
```diff
@@ -3,11 +3,150 @@
...
+The examples show how a cli application could use the hydra client for interaction with an API.
+
+## Find API entrypoint for website
+
+Hydra has a special mechanism to discover API entrypoints.
+
+```
+hydra.findEntrypoint(website)
+print hydra.representation 
+// the API entrypoint is now the current representation
```
dschulten: I had started like that and found the resulting code difficult to handle. We could certainly rewrite hydra to return the representation, but let us consider a bit first.

The question is, should the application that uses the hydra client maintain the application state or should the hydra client do that?

"Not easy to understand" might indicate that it is a novel approach. Current http clients "retrieve stuff". The hydra client is different. It is an application, and you drive it by interacting with it.

drafts/client-interface.md
```diff
@@ -3,11 +3,150 @@
...
+This assumes that the client is aware of the uniform interface of the underlying protocol. 
+In the example below the client knows that it uses the uniform interface of HTTP.
+```
+hydra.setLocation(apiurl);
+// apiurl is now the current representation
+
+// we want to retrieve or download the menu of the food establishment
+menu = hydra.controls["http://schema.org/hasMenu"]
+if(menu && menu.methods["GET"])
+    hydra.execute("GET", menu.href)
```
dschulten: You can't be sure that the hasMenu link supports GET. Or does hasLinkOfType mean that only GET affordances are eligible?
If so, should there be hasPutOfType, hasPostOfType, hasPatchOfType etc.?

drafts/client-interface.md
dschulten: Phew, I just felt obliged to do something with the DownloadAction since you added it in the original draft. I'll delete that section for now, OK.

drafts/client-interface.md
dschulten: Let me argue that  we should make hydraClient a browser of sorts. If you work with an API, you visit several representations until you finally reach your goal. If you have moved to a different representation, you are usually not interested in the previous representations anymore. Then why should the user of the hydraClient hold references to all those representations?

Can you please illustrate your point that there may be multiple branches or error handling which are difficult to handle if there is only one current representation and the hydraClient holds it?

drafts/client-interface.md
dschulten: In my experience many people mistake UriTemplates as equivalent of a GET form in the browser, i.e. they think the template is a recipe to create a link for a GET method, primarily with '?' query parameters. That is why I stress that a UriTemplate describes a valid URI space.
I can change the title, but I would keep the quote from the RFC, if that is OK.

drafts/client-interface.md
dschulten: Multiple links are a good point. How should we enable the user of the hydraClient to identify the link she is looking for? There can be repeated links on collection items. There also can be links with the same predicate in totally unrelated contexts, e.g. two search links on two different collections. Would it be sufficient if a control knows to which item it is directly attached?

dschulten: Otherwise, I am all for syntactic sugar once we have the barebone functionality in place. I totally get that  a simple GET should be simple (pun intended)

tpluscode:
That's more how I started with heracles. You can start with any resource URL and get a Hydra resource object with `hydra:stuff` mapped to nice JS getters. It will be possible to define custom extension so that it won't be necessary to work with full URIs name.

Here's the direction I want to go

```
import {Hydra} from 'heracles';

Hydra.loadResource('http://example.com/user/tpluscode')
  .then(res => {
    // contains the entrypoint, supported classes and operations, and other Api Doc stuff.
    var apiDocs = res.apiDocumentation;
    
    // same as res['@id']
    var id = res.id; 
   
    // let's assume the operation is available
   var op = res.operations["http://example.com/vocab#updateName"];

    return op.invoke({
        'name': 'Tomasz'
    });
  }).then(doSomethingWithTheOperationResult);
```

As you see there is no state. Or rather, there are only Hydra resource

On the other hand it now dawns on me that I intend to create a Hydra library and not a client right?

drafts/client-interface.md
```diff
@@ -1,13 +1,133 @@
...
+
+The following code assumes that the client is aware of the uniform interface of the underlying protocol. 
+In the example below the client knows that it uses the uniform interface of HTTP.
+```
+hydraClient.setLocation(apiurl);
+// apiurl is now the current representation
+
+// we want to retrieve or download the menu of the food establishment
+menu = hydraClient.controls["http://schema.org/hasMenu"]
+if(menu && menu.methods["GET"])
```
tpluscode: I'm interested why the consumer should care whether it's a GET or another method? Wouldn't it be better to have it hidden behind the operation's interface. At least for the purpose of invoking/executing the operation.

RubenVerborgh:
I find the interaction with the "hypothetical hydra object" a bit strange; it's a weird abstraction from a browser (minus tabs etc.). What are reasons to prefer this over a resource-based abstraction, where one resource provides access to others?

dschulten:
Well it is clear that so far no one shares my idea of a hydra client interface. I have no problem with that. I want us to make progress here, not to defend my ideas. 

My focus is to get a reference client implementation so that we can *verify the Hydra spec* and find out what is really needed. I do not think it is essential for the purpose of verification of the Hydra vocab which metaphor the client abstraction uses or which method signatures it has. We are again splitting hairs, and I am not going to waste any more time on that.

How about someone else spells out his idea of a client interface? Or can we just adopt heracles as the reference client? 

Can we please decide *anything* which moves us forward? I am extremely frustrated with the way the group un-collaborates. No offense intended.

asbjornu:
I really like the idea behind this. Having a reference implementation as an informational (non-normative) appendix in the spec would be great. But why does this reference implementation have to be pseudo code? Can't it instead be a real-world client implementation?

One way to do this would be to move [Heracles](https://github.com/wikibus/heracles) to the @HydraCG organisation and make it work as a foundation for our client-side endeavours. This client could then be offered up as a way to verify server implementations, sort of like "Atom Application Exerciser" worked back when we developed the Atom Publishing Protocol.

RubenVerborgh:
> Well it is clear that so far no one shares my idea of a hydra client interface. I have no problem with that.

It might very well be the best idea, so please explain it that we all are able to understand.

> I want us to make progress here

Then it is very important that we all understand :-)

So could you please give my question another take?

>> What are reasons to prefer this over a resource-based abstraction, where one resource provides access to others?

RubenVerborgh:
> Can we please decide anything which moves us forward? I am extremely frustrated with the way the group un-collaborates. No offense intended.

BTW I share your concern; my architecture e-mail still remains unanswered.

I think having a client is a great idea—perhaps even a very good starting point for a 1.0 spec.
But in that sense, the client spec proposed here can also be seen as "splitting hairs", as it zooms in on very detailed aspects before getting the bigger picture straight.

How about we (= whoever wants to contribute to Hydra) have a conf call to sync on this?

dschulten:
@tpluscode is it an option to let Heracles be the reference client? If so, can we move it, or a fork of it, to the HydraCg organization?

tpluscode:
> is it an option to let Heracles be the reference client?

I don't see why not, in principle. But I hope you didn't get my comment the wrong way. I simply meant that I feel that the kind of client @RubenVerborgh was describing is more natural to me. It is more of an experiment in that I flesh it out as I go. I first added some stuff, then tried to use it, and then refactored. The effect may not be 100% how I'd like it and some code is not pretty ;)

> But why does this reference implementation have to be pseudo code? 

I kind of think it's a good idea up to some point. For now I'm actually in favor of a language-agnostic draft of how we agree such a Hydra client would be implemented and APIs consumed. This way we can at least avoid some intricacies of processing JSON-LD. At least for now. It could also help us decide on some pieces of Hydra by investigating whet client can do rather than how we imagine the vocabulary. I may have some insight after starting with Heracles and so would other people that already started some client implementation. Some names include @antoniogarrote, @bergos, @dunglas, @pchampin, @Mec-iS. Can we get you guys on board?

> How about we have a conf call to sync on this?

Yes, please

RubenVerborgh:
>> How about we have a conf call to sync on this?
> Yes, please

I'll await some more comments, but plan to send out such a mail this week on the Hydra mailing list.

dunglas:
Hi guys, I'll announce a parser for the Hydra documentation, an admin generator (using Admin On Rest) and a CRUD generator (scaffolding an app built with React and Redux) Friday during the Symfony Live conference.

The code is already publicly available:

* The parser: https://github.com/dunglas/api-doc-parser
* The admin: https://github.com/api-platform/admin
* The CRUD generator: https://github.com/api-platform/generate-crud

Both tools are designed to rely on an intermediate representation an to be able to support other API docs format (Swagger, API Blueprint...). Hypermedia features of Hydra aren't 100% supported yet but I think it's a good start. It would be nice to have some rock solid JS libraries to create apps, it will help onboarding new developers.

drafts/client-interface.md
lanthaler: It's probably only because I intended this to be a lower-level thing than what you proposed. I had envisioned the client library to be the thing that is being used by apps like the one you propose in this document. Could we perhaps document your Hydra Browser as a layer implemented on top of a more low-level Hydra Client? I'm completely fine with doing so in this file as I don't think we know yet which alternative is better

drafts/client-interface.md
lanthaler: > Or does hasLinkOfType mean that only GET affordances are eligible?

Yes, that was my intention behind that clunky name :-)

> If so, should there be hasPutOfType, hasPostOfType, hasPatchOfType etc.?

I'd abstract it from HTTP. There are safe transitions (i.e., simple links or GET) and unsafe ones (operations - let's squint for a bit on the fact that operations might be safe as well for now)

> Multiple links are a good point. How should we enable the user of the hydraClient to identify the link she is looking for?

That's up to the app but I imagine it would need to look at properties of the link. If there are none, it would likely consider them as equivalent and follow both.

drafts/client-interface.md
lanthaler: > Let me argue that we should make hydraClient a browser of sorts. If you work with an API, you visit several representations until you finally reach your goal.

For me, a browser is an app. I envisioned the Hydra client to be something lower level. There's no better or worse, they are just different levels of abstraction.

> If you have moved to a different representation, you are usually not interested in the previous representations anymore. Then why should the user of the hydraClient hold references to all those representations?

I'd argue that depends entirely on the app, i.e., the goal you are trying to achieve

> Can you please illustrate your point that there may be multiple branches or error handling which are difficult to handle if there is only one current representation and the hydraClient holds it?

Imagine I pass `hydra` to a method which in turn passes it to another method and then I continue. What's the state of `hydra`? I'd need to introspect it to be sure. How should it behave if an operation fails (network error e.g.)? Should it stay at the previous state or move internally to an error state?

drafts/client-interface.md
lanthaler: Fine with me

lanthaler:
Wow, I didn't intend to side-track this PR so much. My reaction (and perhaps the one of others too) was solely triggered by the fact that I expected the Hydra client library to be something lower level. I'd suggest we call that out explicitly in the doc, merge this PR and iterate.

Thanks a lot @dschulten for taking the lead on this and please bear with us. The amount of feedback you received shows that people definitely care about your work.

tpluscode:
I agree. Both client library and Hydra Client (capitalized) have their place.

drafts/client-interface.md
tpluscode: > I'd abstract it from HTTP

Yes, I think I'd prefer a high-level client to operation on abstracted model. I even't prefer that in a low level lib. 

Traditionally we called safe and unsafe transitions Links and Operations (Actions, arguably). Are we sticking to this naming?

dschulten:
@lanthaler the original version guided me to believe that the document should not just outline an interface but show a pseudocode application which *makes use* of the hydra client library:

## Retrieve a resource
    resource = hydra.download(url)

I see an application there which uses the hydra client, and that approach makes sense because it allows you to think about the interface from the caller's perspective. The PR in all its shortness goes full circle in that it tries to cover all cases where the caller needs to work with server expectations: uri templates, headers and request bodies.

If somebody else would rather write a totally different initial PR and actually tasks himself to do it very soon, let's wait. Otherwise, we could continue with this PR and someone rewrites it so that `hydraClient` is only a starting point and the caller uses a representation abstraction for all further interaction.

Obviously it is not a good mode of operation if everybody tells me they had a very different idea, asking me to bring their idea to life - accompanied by jeers and catcalls ;-) Or at least that is how it felt on my end.

lanthaler:
> @lanthaler the original version guided me to believe that the document should not just outline an interface but show a pseudocode application which makes use of the hydra client library:

Yes, I indeed wanted to show how the client looks like in use. It doesn't need to be a full application. Some simple use cases (retrieve a resource, retrieve all items from a collection, execute an operation, etc.) should be enough. It helps to understand the usability of the interface.

> If somebody else would rather write a totally different initial PR and actually tasks himself to do it very soon, let's wait.

No, we definitely shouldn't do that IMO :-)


> Obviously it is not a good mode of operation if everybody tells me they had a very different idea, asking me to bring their idea to life - accompanied by jeers and catcalls ;-) Or at least that is how it felt on my end.

Well, we will need to have some discussions at some point. Since the group decided to move most of the work to GitHub, a PR feels like a good opportunity to have such a discussion. Don't see it as criticism but as constructive feedback. It's your PR, so feel free to push back, address it by leaving a note or todo in the doc or something else. I'm actually really happy to see such lively discussions again but I can imagine how it may have felt on your end.

So, to move forward I'd suggest to go through the comments and address them in whatever way you feel is appropriate. Adding a section "open questions" or something along those lines would work quite well for lots of them given that this is a working doc. We can then merge this and someone else can pick up one of the questions or make changes to the doc that we will then discuss separately.

Does that sounds like a reasonable way forward?

RubenVerborgh:
My two cents would be that:
- This is too substantial to discuss in a pull request. We'd really need a repo for a client, so we can have a structured discussion.
- We need a phone conf to discuss this and more. Bringing this to the mailing list now.

drafts/client-interface.md
dschulten: If we make the hydraClient stateless and ask the user of the hydraClient to manage the application state of the API by holding the returned representation, where do we put the ApiDocumentation? Should the user of the hydraClient hold that, too? Or does every representation pass on the ApiDocumentation to the next representation?

lanthaler: The resource that is returned could have a method to get the link to the ApiDocumentation.

Dietrich, what do you think about the proposal 

> to go through the comments and address them in whatever way you feel is appropriate. Adding a section "open questions" or something along those lines would work quite well for lots of them given that this is a working doc. We can then merge this and someone else can pick up one of the questions or make changes to the doc that we will then discuss separately.

I wouldn't like this to get stuck. Your approach (top-down) is completely fine. We don't need to work bottom-up.

PR #113, opened 2017-04-19
Let's consider first an entry-point use case of the Hydra client.

alien-mcl:
Please find the initial version of the entry-point fetch use case for the Hydra client.

### DIFF
--- a/drafts/use-cases/0.intro.md
+++ b/drafts/use-cases/0.intro.md
@@ -1,0 +1,93 @@
+# Uses cases for Hydra Core Vocabulary client
+
+Here you can find some of the use cases to consider while developing a reference client implementation and improve Hydra Core Vocabulary specification.
+
+To create a context of the whole task, these use-cases are related to the [EventAPI](https://github.com/lanthaler/EventApiDemo) and a calendar-like UI web application.
+
+Each scenario is build using several sections:
+
+- Story - a short _As a... I want... so..._ story introducing a general requirement
+- Usage - a pseudo-code (JavaScript/TypeScript like) of the imaginary client
+- Details - detailed description
+- Considerations - some additional considerations on several aspects
+
+## Assumptions
+
+For simplification, some assumptions are made so the main topic of each of the use-cases are not distracted with side details.
+
+These would be:
+
+- client-server communication uses JSON-LD RDF serialization using HTTP headers:
+  Content-Type: application/ld+json for server responses
+  Accept: application/ld+json client requests
+- JSON-LD context is described later on in this document and is referenced in HTTP header:
+  Link: </events/events.jsonld>; rel="http://www.w3.org/ns/json-ld#context"; type="application/ld+json"
+- main URL used is _/api_
+
+### JSON-LD context
+```javascript
+{
+    "hydra": "http://www.w3.org/ns/hydra/core#",
+    "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
+    "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
+    "xsd": "http://www.w3.org/2001/XMLSchema#",
+    "owl": "http://www.w3.org/2002/07/owl#",
+    "schema": "http://schema.org/",
+    "title": {
+        "@id": "hydra:title",
+        "@type": "xsd:string"
+    },
+    "operations": {
+        "@id": "hydra:operation",
+        "@type": "hydra:Operation"
+    },
+    "httpMethod": {
+        "@id": "hydra:method",
+        "@type": "xsd:string"
+    },
+    "returns": {
+        "@id": "hydra:returns",
+        "@type": "hydra:Class"
+    },
+    "expects": {
+        "@id": "hydra:expects",
+        "@type": "hydra:Class"
+    },
+    "supportedClasses": {
+        "@id": "hydra:supportedClass",
+        "@type": "hydra:Class"
+    },
+    "supportedProperties": {
+        "@id": "hydra:supportedProperty",
+        "@type": "@id"
+    },
+    "property": {
+        "@id": "hydra:property",
+        "@type": "rdf:Property"
+    },
+    "members": {
+        "@id": "hydra:member",
+        "@type": "@id"
+    },
+    "totalItems": {
+        "@id": "hydra:totalItems",
+        "@type": "xsd:int"
+    },
+    "eventName": {
+        "@id": "schema:name",
+        "@type": "xsd:string"
+    },
+    "eventDescription": {
+        "@id": "schema:description",
+        "@type": "xsd:string"
+    },
+    "startDate": {
+        "@id": "schema:startDate",
+        "@type": "xsd:dateTime"
+    },
+    "endDate": {
+        "@id": "schema:endDate",
+        "@type": "xsd:dateTime"
+    }
+}
+```
--- a/drafts/use-cases/0.intro.md
+++ b/drafts/use-cases/0.intro.md
@@ -20,74 +20,82 @@ These would be:
 - client-server communication uses JSON-LD RDF serialization using HTTP headers:
   Content-Type: application/ld+json for server responses
   Accept: application/ld+json client requests
-- JSON-LD context is described later on in this document and is referenced in HTTP header:
-  Link: </events/events.jsonld>; rel="http://www.w3.org/ns/json-ld#context"; type="application/ld+json"
-- main URL used is _/api_
+- JSON-LD context is described later on in this document and is referenced in further example
+  payloads with this property:
+  ```javascript
+  {
+      "@context": "/api/context.jsonld"
+      // further example payload goes here
+  }
+  ```
+- to simplify, the host was dropped from all Urls. We can assume an example like http://example.com/.
 
 ### JSON-LD context
 ```javascript
 {
-    "hydra": "http://www.w3.org/ns/hydra/core#",
-    "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
-    "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
-    "xsd": "http://www.w3.org/2001/XMLSchema#",
-    "owl": "http://www.w3.org/2002/07/owl#",
-    "schema": "http://schema.org/",
-    "title": {
-        "@id": "hydra:title",
-        "@type": "xsd:string"
-    },
-    "operations": {
-        "@id": "hydra:operation",
-        "@type": "hydra:Operation"
-    },
-    "httpMethod": {
-        "@id": "hydra:method",
-        "@type": "xsd:string"
-    },
-    "returns": {
-        "@id": "hydra:returns",
-        "@type": "hydra:Class"
-    },
-    "expects": {
-        "@id": "hydra:expects",
-        "@type": "hydra:Class"
-    },
-    "supportedClasses": {
-        "@id": "hydra:supportedClass",
-        "@type": "hydra:Class"
-    },
-    "supportedProperties": {
-        "@id": "hydra:supportedProperty",
-        "@type": "@id"
-    },
-    "property": {
-        "@id": "hydra:property",
-        "@type": "rdf:Property"
-    },
-    "members": {
-        "@id": "hydra:member",
-        "@type": "@id"
-    },
-    "totalItems": {
-        "@id": "hydra:totalItems",
-        "@type": "xsd:int"
-    },
-    "eventName": {
-        "@id": "schema:name",
-        "@type": "xsd:string"
-    },
-    "eventDescription": {
-        "@id": "schema:description",
-        "@type": "xsd:string"
-    },
-    "startDate": {
-        "@id": "schema:startDate",
-        "@type": "xsd:dateTime"
-    },
-    "endDate": {
-        "@id": "schema:endDate",
-        "@type": "xsd:dateTime"
+    "@context": {
+        "hydra": "http://www.w3.org/ns/hydra/core#",
+        "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
+        "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
+        "xsd": "http://www.w3.org/2001/XMLSchema#",
+        "owl": "http://www.w3.org/2002/07/owl#",
+        "schema": "http://schema.org/",
+        "title": {
+            "@id": "hydra:title",
+            "@type": "xsd:string"
+        },
+        "operations": {
+            "@id": "hydra:operation",
+            "@type": "hydra:Operation"
+        },
+        "method": {
+            "@id": "hydra:method",
+            "@type": "xsd:string"
+        },
+        "returns": {
+            "@id": "hydra:returns",
+            "@type": "hydra:Class"
+        },
+        "expects": {
+            "@id": "hydra:expects",
+            "@type": "hydra:Class"
+        },
+        "supportedClasses": {
+            "@id": "hydra:supportedClass",
+            "@type": "hydra:Class"
+        },
+        "supportedProperties": {
+            "@id": "hydra:supportedProperty",
+            "@type": "@id"
+        },
+        "property": {
+            "@id": "hydra:property",
+            "@type": "rdf:Property"
+        },
+        "members": {
+            "@id": "hydra:member",
+            "@type": "@id"
+        },
+        "totalItems": {
+            "@id": "hydra:totalItems",
+            "@type": "xsd:int"
+        },
+        "eventName": {
+            "@id": "schema:name",
+            "@type": "xsd:string"
+        },
+        "eventDescription": {
+            "@id": "schema:description",
+            "@type": "xsd:string"
+        },
+        "startDate": {
+            "@id": "schema:startDate",
+            "@type": "xsd:dateTime"
+        },
+        "endDate": {
+            "@id": "schema:endDate",
+            "@type": "xsd:dateTime"
+        }
     }
 }
 ```
--- a/drafts/use-cases/1.1.security-considerations.md
+++ b/drafts/use-cases/1.1.security-considerations.md
@@ -1,0 +1,40 @@
+### Security considerations
+
+#### JSON array root security issues
+Due to security issues with JSON arrays which can be exploited in some circumstances, 
+there may be a need to avoid pure arrays in favour of a JSON object.
+In the scenario above, server returned a single object, which won't cause issues.
+But it is possible to return multiple resources, resulting in an JSON array returned 
+(server may not use any JSON-LD context or it may be completely different from the one used by the client).
+There are several possibilities here. One of them would be to use a *@graph* notation, i.e.:
+
+```javascript
+{
+    "@graph": [
+        {
+            "@id": "/api",
+            "operations": [
+                {
+                    "@id": "/api/events",
+                    "title": "Events",
+                    "httpMethod": "GET",
+                    "returns": "hydra:Collection"
+                },
+                {
+                    "@id": "/api/events",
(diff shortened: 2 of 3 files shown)


drafts/use-cases/0.intro.md
```diff
@@ -0,0 +1,93 @@
...
+## Assumptions
+
+For simplification, some assumptions are made so the main topic of each of the use-cases are not distracted with side details.
+
+These would be:
+
+- client-server communication uses JSON-LD RDF serialization using HTTP headers:
+  Content-Type: application/ld+json for server responses
+  Accept: application/ld+json client requests
+- JSON-LD context is described later on in this document and is referenced in HTTP header:
```
lanthaler: Referencing a context in the header [isn't supported](http://www.w3.org/TR/json-ld/#interpreting-json-as-json-ld) if the content-type is `application/ld+json`. Maybe just say it's implied or, if probably better, just add that one-liner referencing it.

drafts/use-cases/0.intro.md
```diff
@@ -0,0 +1,93 @@
...
+  Link: </events/events.jsonld>; rel="http://www.w3.org/ns/json-ld#context"; type="application/ld+json"
+- main URL used is _/api_
+
+### JSON-LD context
+```javascript
+{
+    "hydra": "http://www.w3.org/ns/hydra/core#",
```
lanthaler: This context is invalid. These term definitions need to be inside of a `@context` property.

drafts/use-cases/1.entry-point.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Story
+As an application user
+I want my application to consume an entry point for the Web API by providing a base URL
+So I can start managing my events.
+
+### Usage
+```javascript
+var client = new HydraClient();
+var data = client.get("/api");
```
lanthaler: This should probably an absolute URL 

    client.get("http://www.example.com/api"); 

Even more interesting, or at least worth mentioning as well, would be 

    client.get("http://www.example.com/"); 

which then uses the Hydra API documentation (`hydra:entrypoint`) to find the API's `http://www.example.com/api`.

drafts/use-cases/1.entry-point.md
```diff
@@ -0,0 +1,108 @@
...
+for (var control of data.controls) {
+    // do something with the _control_
+}
+```
+
+### Details
+We can imaging a situation, when an application has a base URL provided, i.e. in a configuration or in the input field.
+Once provided, I'd expect the application to connect to the entry-point of the Web API and download initial
```
lanthaler: Nit: This `I'd` should probably be changed as all of this should reflect the groups consensus eventually and not a personal opinion.

drafts/use-cases/1.entry-point.md
```diff
@@ -0,0 +1,108 @@
...
+    "@id": "/api",
+    "operations": [
+        {
+            "@id": "/api/events",
+            "title": "Events",
+            "httpMethod": "GET",
+            "returns": "hydra:Collection"
+        },
+        {
+            "@id": "/api/events",
```
lanthaler: You shouldn't give the same identifier to two different operations. Perhaps you intended to set the target? That's what's at the other end of `hydra:operation`, in this case `/api`

drafts/use-cases/0.intro.md
```diff
@@ -0,0 +1,93 @@
...
+    "schema": "http://schema.org/",
+    "title": {
+        "@id": "hydra:title",
+        "@type": "xsd:string"
+    },
+    "operations": {
+        "@id": "hydra:operation",
+        "@type": "hydra:Operation"
+    },
+    "httpMethod": {
```
lanthaler: Do you want to propose to rename `method` to `httpMethod`? Otherwise I wouldn't remap it (things like plurals as in operations are fine though, even more so if you would also set `"@container": "@set"` as that would ensure values of such properties remain arrays when compacting the JSON-LD)

drafts/use-cases/1.entry-point.md
```diff
@@ -0,0 +1,108 @@
...
+            "title": "Create new event",
+            "httpMethod": "POST",
+            "expects": "schema:Event"
+        }
+    ]
+}
+```
+
+With this details, client application is aware of two possible operations that can be made from that point:
```
lanthaler: Nit: *these* details

drafts/use-cases/1.entry-point.md
```diff
@@ -0,0 +1,108 @@
...
+
+- GET events
+- POST an event
+
+We can imagine a UI application with some fancy logic that would takes all
+GET operations returning _hydra:Collection_ and displays them in a main menu.
+
+### Considerations
+
+#### JSON array root security issues
```
lanthaler: This is a good point but I would move it to a separate document which collects security considerations.

drafts/use-cases/1.entry-point.md
```diff
@@ -0,0 +1,108 @@
...
+        }
+    ]
+}
+```
+
+Using default graph seems reasonable, but implies some additional, possibly unwanted,
+logic of the graph itself.
+We could consider another approach or somehow standardize the graph name (if necessary).
+
+#### Reasoning capabilities
```
lanthaler: Good that you brought this up as most people unfamiliar with RDF wouldn't know that. I'd turn it around though and explicitly include the `@type` in all examples and describe in this section here that it would be possible to omit that, iff clients do RDFS entailment.

lanthaler:
Kudos! This is fantastic work

drafts/use-cases/1.entry-point.md
tpluscode: Interesting. I never though about that. So you suggest that using a hydra client to `GET` the documentation URI should give you state of the entrypoint? If so, I'd mention that in the `ApiDocumentation` page

drafts/use-cases/1.entry-point.md
tpluscode: Not just a nit really. Let's keep these documents impersonal

drafts/use-cases/1.entry-point.md
```diff
@@ -0,0 +1,108 @@
...
+- GET events
+- POST an event
+
+We can imagine a UI application with some fancy logic that would takes all
+GET operations returning _hydra:Collection_ and displays them in a main menu.
+
+### Considerations
+
+#### JSON array root security issues
+Due to security issues with JSON arrays which can be exploited in some circumstances,
```
tpluscode: It would be nice to have some link or elaboration about possible issues.

drafts/use-cases/1.entry-point.md
```diff
@@ -0,0 +1,108 @@
...
+In OWL environment, this is fully legal as the reasoning process would extend the
+resource _/events_ with additional statement (in Turtle):
+
+```
+</api/events> a hydra:Operation .
+```
+
+This is due to fact that Hydra Core Vocabulary provides an _rdfs:domain_ for _hydra:method_ .
+Unfortunately, many clients, especially browser based, won't use that process of reasoning,
+thus an explicit statement should appear so the client can easily discover all the operations.
```
tpluscode: And what if we provided guidelines/hints for "manual" reasoning. Akin to rule-based inferences? That way it should theoretically be possible for someone to implement a generic client with little-to-none RDF/OWL knowledge.

On the other hand, while implementing heracles, this quickly got out of hand and did in fact introduce actual triple processing...

drafts/use-cases/1.entry-point.md
```diff
@@ -0,0 +1,108 @@
...
+```javascript
+var client = new HydraClient();
+var data = client.get("/api");
+for (var control of data.controls) {
+    // do something with the _control_
+}
+```
+
+### Details
+We can imaging a situation, when an application has a base URL provided, i.e. in a configuration or in the input field.
```
tpluscode: Typo: We can *imagine*

drafts/use-cases/1.entry-point.md
lanthaler: Sorry, that was a copy-paste error which I just corrected. What I meant was something like

    client.get("http://www.example.com/").getEntrypoint(); 

or

    client.get("http://www.example.com/").getApiDocumentation().getEntrypoint();

The client would look for the Hydra HTTP header, fetch the API documentation, and return the entrypoint.

drafts/use-cases/0.intro.md
asbjornu: Perhaps just rewrite the sentence to:

> The JSON-LD context is described later in this document and **can be** referenced in an HTTP header as such:

drafts/use-cases/1.entry-point.md
asbjornu: Since this entry point discovery is something that needs to be performed often unless cached, should there be any mention of caching here? It surely applies to all URIs consumed by the Hydra Client, but the entry point is a contention point.

Also, if we mention caching, in what detail should we go with regards to cache lifetime, invalidation, etc.? Just refer to RFC 7234 or what?

drafts/use-cases/1.entry-point.md
asbjornu: I agree. There's a few personal pronouns used across all documents that should be replaced.

drafts/use-cases/1.entry-point.md
asbjornu: It's not only a security issue returing a JSON array; it's an extensibility issue. Not having an object at the representation root means there's no way to extend without breakage. This should be a design principle in Hydra, imho.

drafts/use-cases/1.entry-point.md
```diff
@@ -0,0 +1,108 @@
...
+}
+```
+
+Using default graph seems reasonable, but implies some additional, possibly unwanted,
+logic of the graph itself.
+We could consider another approach or somehow standardize the graph name (if necessary).
+
+#### Reasoning capabilities
+Another thing is how far the client should be pushed into the reasoning process.
+In the example above, only _hydra:method_ and _hydra:expects_ predicates were used.
```
asbjornu: Instead of `_`, can you perhaps wrap all of these terms in backticks (`)?

drafts/use-cases/1.entry-point.md
asbjornu: Can the clients **not** do entailment? Wouldn't that be extremely limiting in what the API is able to express?

drafts/use-cases/1.entry-point.md
asbjornu: > That way it should theoretically be possible for someone to implement a generic client with little-to-none RDF/OWL knowledge.

That's what I hope, or else I won't be able to implement a Hydra Client at all. 😄 

> On the other hand, while implementing heracles, this quickly got out of hand and did in fact introduce actual triple processing...

Isn't that something that would be required either way? I suppose all of the JSON-LD and RDF massage would have to be performed by a specialised library that the Hydra Client builds on top of?

drafts/use-cases/1.entry-point.md
lanthaler: Yeah, good point. I'd probably move all of those discussions to a separate Performance document though. We can then simply refer it here. Something like

> In typical usage scenarios, the API documentation needs to be used frequently and should thus [be cached] (link to Performance).

drafts/use-cases/0.intro.md
alien-mcl: OK. I'll rewrite it. Still, I was wondering why this is not supported - changing one value in an HTTP header enforces incomparably more complex changes on both client and server side.

drafts/use-cases/0.intro.md
alien-mcl: I'll have that fixed.

drafts/use-cases/1.entry-point.md
alien-mcl: But what if the client gets _http://www.example.com/api_ - which is already an entry point. This way you won't fetch any API documentation headers (I wouldn't expect those here).
From my experience APIs (_http://www.example.com/api_ in this example) are usually added to the existing site (_http://www.example.com/_) and the main Url doesn't care about the API.
I believe we shall have these scenarios covered in the spec.

alien-mcl: One more thing - I wouldn't force the client to get the entry point via API documentation. I though both are independent to each other.

drafts/use-cases/1.entry-point.md
alien-mcl: Hmm - I acted from an product owner somehow :). Anyway, I'll make it impersonal as you suggested.

drafts/use-cases/1.entry-point.md
alien-mcl: True that. I'll have that fixed

drafts/use-cases/0.intro.md
alien-mcl: That was not my intention. I though that would be more readable. I'll retract that change.

drafts/use-cases/1.entry-point.md
alien-mcl: OK

drafts/use-cases/1.entry-point.md
alien-mcl: That was my point. Specification should explicitly describe both how the hypermedia controls/API documentation should be expressed and what client capabilities (if any) must exist. Currently it's pure RDF over JSON and all RDFS/OWL capabilities may be required.

drafts/use-cases/1.entry-point.md
alien-mcl: I'm pleased that my concerns (the main driving force of putting these paragraphs here) are shared.
I had similar issues while implementing (URSA)[https://github.com/alien-mcl/URSA]. In order to fully express all the stuff I needed I had to use not-that-simple constructs. I'd expect the spec to explicitly define these kind of rules. OWA does not make things easier for simple clients

drafts/use-cases/1.entry-point.md
alien-mcl: True that

drafts/use-cases/1.entry-point.md
alien-mcl: I've change that to be impersonal.

drafts/use-cases/1.entry-point.md
alien-mcl: I'll use asterisk instead - in rich markdown views it'll use either italics or bold, in plain text it should still be readable.

drafts/use-cases/1.entry-point.md
lanthaler: > But what if the client gets http://www.example.com/api - which is already an entry point. This way you won't fetch any API documentation headers (I wouldn't expect those here).

Why not? If an API documentation exists, it should be referenced form any resources that is part of the API... and, if possible, also from every other resource the API publisher controls.

> From my experience APIs (http://www.example.com/api in this example) are usually added to the existing site (http://www.example.com/) and the main Url doesn't care about the API.

Yeah, that's the trickier case. It won't always be practical to do that but we should document that it is possible and enables automatic discovery.

> I wouldn't force the client to get the entry point via API documentation. I though both are independent to each other.

Of course not. It's important though to show that it is possible. The scenario I'm thinking about is that you for example browser a website. You get curious whether it would be possible to get data automatically. If the HTTP header is included, the browser (or an extension) can immediately tell you that an API is available without you having to look for it (on maybe a different subdomain).

drafts/use-cases/0.intro.md
lanthaler: That feature is an escape hatch for existing APIs that can't change the content-type. An HTTP header is difficult to discover and quite separated from the document itself. We decided that if someone uses the `application/ld+json` media type, all information should be contained in the document and not hidden in headers (which makes copy-pasting for instance impossible).

drafts/use-cases/1.entry-point.md
lanthaler: Yep, the spec should make that clear.

drafts/use-cases/1.entry-point.md
alien-mcl: >Why not? If an API documentation exists, it should be referenced form any resources that is part of the API...
>and, if possible, also from every other resource the API publisher controls.
Maybe this requirement should be added to the spec? I believe we had a short discussion on expressing what client must, should and may understand.

drafts/use-cases/1.entry-point.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Story
+As an application user
+I want my application to consume an entry point for the Web API by providing a base URL
+So I can start managing my events.
+
+### Usage
+```javascript
+var client = new HydraClient();
+var entryPoint = client.get("http://example.com/").getEntryPoint();
```
elf-pavlik: This `client.get()` looks pretty weird as synchronous method, how about using promises in similar way to https://fetch.spec.whatwg.org/ ?

```js
var client = new HydraClient()
client.get("http://example.com/")
  .then(response => response.getEntryPoint())
  .then(entryPoint => {
    for (var operation of entryPoint.operations) {
      // do something with the *operation*
    }
  }).catch(err => { /* handle errors */ })
```

tpluscode: I don't think it's all that important. Despite the snippet is marked as JS, we should treat any examples here as pseudo-code.

alien-mcl: Oh, exactly. I didn't want to focus on any specific solution in this matter as there are plenty of (mentioned promises, ES17 async/await, RX, you name it). Indeed I'm using JS but duw ro fact it is juat simpler to use an existing syntax familiar for most of us than to write some abstract commands.

elf-pavlik: Ok, we can always later on update all the code snippets via separate PR

drafts/use-cases/0.intro.md
```diff
@@ -0,0 +1,101 @@
...
+        "xsd": "http://www.w3.org/2001/XMLSchema#",
+        "owl": "http://www.w3.org/2002/07/owl#",
+        "schema": "http://schema.org/",
+        "title": {
+            "@id": "hydra:title",
+            "@type": "xsd:string"
+        },
+        "operations": {
+            "@id": "hydra:operation",
+            "@type": "hydra:Operation"
```
elf-pavlik: In JSON-LD playground http://tinyurl.com/lqhm342 with this context `"operations": "http://foo.example/bar"` when expanded it seems like a literal (`@value` not `@id`)
```json
[
  {
    "http://www.w3.org/ns/hydra/core#operation": [
      {
        "@type": "http://www.w3.org/ns/hydra/core#Operation",
        "@value": "http://foo.example/bar"
      }
    ]
  }
]

after changing `"@type": "hydra:Operation"` to `"@type": "@id"` it looks like named node again
```json
[
  {
    "http://www.w3.org/ns/hydra/core#operation": [
      {
        "@id": "http://foo.example/bar"
      }
    ]
  }
]
```

drafts/use-cases/1.1.security-considerations.md
```diff
@@ -0,0 +1,40 @@
...
+There are several possibilities here. One of them would be to use a *@graph* notation, i.e.:
+
+```javascript
+{
+    "@graph": [
+        {
+            "@id": "/api",
+            "operations": [
+                {
+                    "@id": "/api/events",
```
elf-pavlik: Given `/api/events` denotes the operation `"@type": "hydra:Operation"`, would some different URI would denote the resource `"@type": "hydra:Collection` on which client can perform this operation? e.g `/data/events`

drafts/use-cases/0.intro.md
lanthaler: @elf-pavlik as agreed in the telecon, I'm going to merge these files now. Could you please file PR to fix this? Thanks

elf-pavlik: PR #120

PR #114, opened 2017-04-19
Let's consider an API documentation fetch use case for the Hydra client.

alien-mcl:
Please find the initial version of the API documentation fetch use case for the Hydra client.


drafts/use-cases/0.intro.md
```diff
@@ -0,0 +1,93 @@
+# Uses cases for Hydra Core Vocabulary client
```
lanthaler: I'll ignore this file here as I just reviewed it in #113

drafts/use-cases/2.api-documentation.md
```diff
@@ -0,0 +1,111 @@
+## API Documentation
+
+### Story
+As an application user
+I want my application to be able to perform some precomputations 
+So it can pre-generate some views and other UI elements to speed things up.
```
lanthaler: Another story might be

As a API consumer
I want to be able to get an overview of the structure and capabilities of a Hydra API
So that I understand how to interact with the API and make those interactions more efficient.

drafts/use-cases/2.api-documentation.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Usage
+```javascript
+var client = new HydraClient();
+var apiDocumentation = client.getApiDocumentation("/api");
```
lanthaler: Should probably be an absolute URL

drafts/use-cases/2.api-documentation.md
```diff
@@ -0,0 +1,111 @@
...
+thus client is expected to store it temporarely at least for the current session.
+
+```
+GET /api?documentation
+
+HTTP 200 OK
+```
+```javascript
+{
+    "@id": "/api?documentation",
```
lanthaler: Should be `/apidoc` or whatever was used in the pseudo-code above. Same for the `GET`

drafts/use-cases/2.api-documentation.md
```diff
@@ -0,0 +1,111 @@
...
+    "supportedClasses": [
+        {
+            "@id": "schema:Event",
```
lanthaler: I wonder whether we should use prefixes for Schema.org everywhere to make it clear what belongs to Hydra and what doesn't!?

drafts/use-cases/2.api-documentation.md
```diff
@@ -0,0 +1,111 @@
...
+            ]
+        }
+    ]
+}
+```
+
+In order to i.e. generate views for each supported class, a complete type definition should be provided.
+Currently, Hydra Core Vocabulary allows that using RDFS/OWL vocabularies.
+
+The documentation should also contain some human readable details like:
```
lanthaler: Should we leave a TODO here to include that in the example?

drafts/use-cases/2.api-documentation.md
```diff
@@ -0,0 +1,111 @@
...
+- operation name
+- operation description
+- parameters and their descriptions
+
+### Considerations
+
+#### Entry-Point/API Documentation fetching precedence
+In the previous scenario, application fetched an entry-point data as it's first action.
+We also could consider an opposite - taking an API documentation if available.
+This would require some standardized mechanism of accesing that documentation.
```
lanthaler: There exists a standardized mechanism for that, see http://www.hydra-cg.com/spec/latest/core/#discovering-a-hydra-powered-web-api

drafts/use-cases/2.api-documentation.md
```diff
@@ -0,0 +1,111 @@
...
+i.e. a main menu with all supported classes.
+
+#### Data structure description
+As already mentioned, Hydra Core Vocabulary allows to embed class description using RDFS/OWL constructs.
+Unfortunately, using those may prove difficult in parsing.
+In the scenario above, we'd need to use _rdfs:range_ predicate to denote the types a given property accepts.
+For simple scenarios it may be just _xsd:string_, but what if server will come out with a union of classes?
+
+#### Embedding class supported operations
+We could embed within each of the supported class details on what can be done with those classes,
```
lanthaler: These are great ideas. Could we mark them up somehow (maybe just as TODO) so that we can find them easily across all docs?

drafts/use-cases/2.api-documentation.md
```diff
@@ -0,0 +1,111 @@
...
+i.e. how can we obtain all instances or how we can create a new one.
+This may be an alternative to creating i.e. main menu in the UI application -
+user will be presented all possibilities, from which some may be denied.
+In case of an entry-point and embedded hypermedia controls I'd expect only those that are allowed in the current state.
+Currently, Hydra Core Vocabulary does not give a clear answer on how such a construct.
+
+#### API documentation and hypermedia controls precedence
+It is unclear on how details obtain from API documentation and current hypermedia controls should be "merged".
```
lanthaler: This is actually not unclear but very well defined due to being based on RDF whose biggest strength is data integration. You just add union the information.

drafts/use-cases/2.api-documentation.md
tpluscode: Agreed. The use case is too vague

drafts/use-cases/2.api-documentation.md
```diff
@@ -0,0 +1,111 @@
...
+var client = new HydraClient();
+var apiDocumentation = client.getApiDocumentation("/api");
+for (var $class of apiDocumentation.supportedClasses) {
+    // generate a partial view for the _$class_
+}
+```
+
+### Details
+The documentation obtain may be a large payload,
+thus client is expected to store it temporarely at least for the current session.
```
tpluscode: Typo: temporarily

drafts/use-cases/2.api-documentation.md
```diff
@@ -0,0 +1,111 @@
...
+Few possibilities here are available:
+- sending an _OPTIONS_ HTTP request to the base URL,
+  but HTTP specification does not allow to cache these responses
+- sending an _GET_ HTTP request to the base URL with some HTTP header;
+  header _Expect_ would be a candidate, but it's not allowed in AJAX communication;
+  another option would be a _Prefer_
+
+Once the API documentation is fetched and processed,
+client would then display some base view with further options,
+i.e. a main menu with all supported classes.
```
tpluscode: > display all supported classes

What would the client do next? I expect the client to work with resource representations. The documentation tells the client how to do it but it is not its main interest.

drafts/use-cases/2.api-documentation.md
```diff
@@ -0,0 +1,111 @@
...
+
+#### Data structure description
+As already mentioned, Hydra Core Vocabulary allows to embed class description using RDFS/OWL constructs.
+Unfortunately, using those may prove difficult in parsing.
+In the scenario above, we'd need to use _rdfs:range_ predicate to denote the types a given property accepts.
+For simple scenarios it may be just _xsd:string_, but what if server will come out with a union of classes?
```
tpluscode: I'm not sure I agree. Sure, using OWL constructs could be a problem but the client won't really need to do inferencing in this particular case. For me description like below would be completely fine

```
{
  "suportedProperties": [
    {
      "@id": "ex:title",
      "rdfs:range": "xsd:string"
    },
    {
      "@id": "ex:spouse",
      "rdfs:range": {
        "owl:unionOf": [ "ex:Class1", "ex:Class2" ]
      }
    }
  ]
}
```

In that the `ex:something` accepts either `ex:Class1` or `ex:Class2`. Note however that I used meaningless names because I couldn't come up with a good example.

Nonetheless, I don't think it's too much to expect the client to understand these semantics.

drafts/use-cases/2.api-documentation.md
tpluscode: We've been on this topic on and off multiple times. I came to accept the union-only solution with a workaround being to shift the variable operations inline or have multiple supported classes which offer a subset of operations.

I haven't got a strong opinion on this anymore but maybe we could look at other API description solutions. Is there any single one that allows one description to cancel out or override another?

drafts/use-cases/0.intro.md
```diff
@@ -0,0 +1,93 @@
...
+
+Here you can find some of the use cases to consider while developing a reference client implementation and improve Hydra Core Vocabulary specification.
```
asbjornu: > ... and improve **the** Hydra Core Vocabulary specification.

drafts/use-cases/0.intro.md
```diff
@@ -0,0 +1,93 @@
...
+
+To create a context of the whole task, these use-cases are related to the [EventAPI](https://github.com/lanthaler/EventApiDemo) and a calendar-like UI web application.
+
+Each scenario is build using several sections:
```
asbjornu: Typo: `built`, not `build`.

drafts/use-cases/2.api-documentation.md
asbjornu: I agree. No need for uppercase `s` in `So`, though.

drafts/use-cases/2.api-documentation.md
```diff
@@ -0,0 +1,111 @@
...
+                    }
+                }
+            ]
+        }
+    ]
+}
+```
+
+In order to i.e. generate views for each supported class, a complete type definition should be provided.
+Currently, Hydra Core Vocabulary allows that using RDFS/OWL vocabularies.
```
asbjornu: > Currently, the Hydra Core Vocabulary allows using RDFS/OWL vocabularies.

drafts/use-cases/2.api-documentation.md
```diff
@@ -0,0 +1,111 @@
...
+
+The documentation should also contain some human readable details like:
+- operation expected type and it's role
+- operation returned type and it's role
```
asbjornu: Typo: `its`, not `it's`.

drafts/use-cases/2.api-documentation.md
```diff
@@ -0,0 +1,111 @@
...
+- operation name
+- operation description
+- parameters and their descriptions
+
+### Considerations
+
+#### Entry-Point/API Documentation fetching precedence
+In the previous scenario, application fetched an entry-point data as it's first action.
```
asbjornu: Typo: `its`, not `it's`.

drafts/use-cases/2.api-documentation.md
```diff
@@ -0,0 +1,111 @@
...
+We also could consider an opposite - taking an API documentation if available.
```
asbjornu: > We also could consider **the** opposite ...

drafts/use-cases/2.api-documentation.md
asbjornu: Yes, this is a contention point in Hydra's current design. I know way too little about RDF to say exactly who or what's to blame, but not being able to **both** override (alter or remove) and extend the hypermedia controls defined in the ApiDocumentation in the representations is a major drawback. I don't think this PR is the place to discuss this issue, but we need to figure out a design that is a bit more flexible than the current, imho.

drafts/use-cases/2.api-documentation.md
```diff
@@ -0,0 +1,111 @@
...
+user will be presented all possibilities, from which some may be denied.
+In case of an entry-point and embedded hypermedia controls I'd expect only those that are allowed in the current state.
+Currently, Hydra Core Vocabulary does not give a clear answer on how such a construct.
+
+#### API documentation and hypermedia controls precedence
+It is unclear on how details obtain from API documentation and current hypermedia controls should be "merged".
+There are few possibilities here:
+
+- hypermedia controls overrides API documentation
+- hypermedia controls extends API documentation
```
asbjornu: I would prefer if we could say "all of the above".

drafts/use-cases/2.api-documentation.md
alien-mcl: OK. I'll change that

drafts/use-cases/2.api-documentation.md
alien-mcl: I'll stick to the convention used in previous comments for #113:
```javascript
var apiDocumentation = client.get("http://example.com/").getApiDocumentation();
```

drafts/use-cases/2.api-documentation.md
alien-mcl: Well, in the light of the comment above this would be fully server driven.
First client would obtain the address of the API documentation from the base Url, and the it would use that address to obtain the API documentation.

drafts/use-cases/2.api-documentation.md
alien-mcl: I tried the opposite - properties of "my" types where clean of any prefixes. Indeed I didn't use that approach for types. I'll make that more consistent.

drafts/use-cases/2.api-documentation.md
alien-mcl: OK - I'll task list to each bullet point below that paragraph.

drafts/use-cases/2.api-documentation.md
alien-mcl: I wasn't aware of it.

drafts/use-cases/2.api-documentation.md
alien-mcl: I'll review all use cases later on and indeed I'll introduce some kind of index of those ideas.

drafts/use-cases/2.api-documentation.md
alien-mcl: I'm happy that both @tpluscode and @asbjornu put @lanthaler's statement into question. Union, or generally OWA, the core principle behind RDF is not really nice to play with. I'd also prefer an explicit way of expressing how the final behavior should be implemented, thus a need for overriding or extending will make developer's life easier.

drafts/use-cases/2.api-documentation.md
alien-mcl: I'd disagree. I already exercised many times applications with main-menus listing a given domain's core resource types like contacts, invoices, etc.

drafts/use-cases/2.api-documentation.md
tpluscode: Still, let's keep the default UNION behavior as-is and leave the overriding for the future.

drafts/use-cases/2.api-documentation.md
alien-mcl: My intention was to shed the light on the fact that the specification neither forbids or forces to use these constructs, but I wouldn't like my client application to have such a strong reasoning capabilities.
Union was an example - what about types define by restrictions?? RDFS has some powerful tools, OWL is even more powerful - I just put to the question whether we need those.

tpluscode: Currently yes, you're right. But we could support more interesting scenarios without requiring `strong reasoning capabilities`. Just a subset of OWL (maybe Hydra-branded?) with simpler semantics.

PR #115, opened 2017-04-19
Let's consider a use case when a Hydra client obtains events.

alien-mcl:
Please find the initial version of the use case when a Hydra client fetches events.

### DIFF
--- a/drafts/use-cases/0.intro.md
+++ b/drafts/use-cases/0.intro.md
@@ -1,0 +1,93 @@
+# Uses cases for Hydra Core Vocabulary client
+
+Here you can find some of the use cases to consider while developing a reference client implementation and improve Hydra Core Vocabulary specification.
+
+To create a context of the whole task, these use-cases are related to the [EventAPI](https://github.com/lanthaler/EventApiDemo) and a calendar-like UI web application.
+
+Each scenario is build using several sections:
+
+- Story - a short _As a... I want... so..._ story introducing a general requirement
+- Usage - a pseudo-code (JavaScript/TypeScript like) of the imaginary client
+- Details - detailed description
+- Considerations - some additional considerations on several aspects
+
+## Assumptions
+
+For simplification, some assumptions are made so the main topic of each of the use-cases are not distracted with side details.
+
+These would be:
+
+- client-server communication uses JSON-LD RDF serialization using HTTP headers:
+  Content-Type: application/ld+json for server responses
+  Accept: application/ld+json client requests
+- JSON-LD context is described later on in this document and is referenced in HTTP header:
+  Link: </events/events.jsonld>; rel="http://www.w3.org/ns/json-ld#context"; type="application/ld+json"
+- main URL used is _/api_
+
+### JSON-LD context
+```javascript
+{
+    "hydra": "http://www.w3.org/ns/hydra/core#",
+    "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
+    "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
+    "xsd": "http://www.w3.org/2001/XMLSchema#",
+    "owl": "http://www.w3.org/2002/07/owl#",
+    "schema": "http://schema.org/",
+    "title": {
+        "@id": "hydra:title",
+        "@type": "xsd:string"
+    },
+    "operations": {
+        "@id": "hydra:operation",
+        "@type": "hydra:Operation"
+    },
+    "httpMethod": {
+        "@id": "hydra:method",
+        "@type": "xsd:string"
+    },
+    "returns": {
+        "@id": "hydra:returns",
+        "@type": "hydra:Class"
+    },
+    "expects": {
+        "@id": "hydra:expects",
+        "@type": "hydra:Class"
+    },
+    "supportedClasses": {
+        "@id": "hydra:supportedClass",
+        "@type": "hydra:Class"
+    },
+    "supportedProperties": {
+        "@id": "hydra:supportedProperty",
+        "@type": "@id"
+    },
+    "property": {
+        "@id": "hydra:property",
+        "@type": "rdf:Property"
+    },
+    "members": {
+        "@id": "hydra:member",
+        "@type": "@id"
+    },
+    "totalItems": {
+        "@id": "hydra:totalItems",
+        "@type": "xsd:int"
+    },
+    "eventName": {
+        "@id": "schema:name",
+        "@type": "xsd:string"
+    },
+    "eventDescription": {
+        "@id": "schema:description",
+        "@type": "xsd:string"
+    },
+    "startDate": {
+        "@id": "schema:startDate",
+        "@type": "xsd:dateTime"
+    },
+    "endDate": {
+        "@id": "schema:endDate",
+        "@type": "xsd:dateTime"
+    }
+}
+```
--- a/drafts/use-cases/3.obtaining-events.md
+++ b/drafts/use-cases/3.obtaining-events.md
@@ -1,9 +1,14 @@
 ## Obtaining events
 
-### Story
-As an application user
-I want to display all my events
-So I can find out what awaits me in the nearest future.
+### Story 
+As an API consumer
+I want to be able to retrieve all members of a collection
+So that I can process them further.
+ 
+### Example
+As an application user 
+I want to display all my events 
+So I can find out what awaits me in the nearest future. 
 
 ### Usage
 ```javascript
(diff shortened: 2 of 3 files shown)


drafts/use-cases/0.intro.md
```diff
@@ -0,0 +1,93 @@
+# Uses cases for Hydra Core Vocabulary client
```
lanthaler: Again, I'll ignore this file as I just reviewed it in #113

drafts/use-cases/3.obtaining-events.md
```diff
@@ -0,0 +1,110 @@
+## Entry point
```
lanthaler: The title should be changed and likely the document's name too. What you are describing here is retrieving members of a collection so maybe you wanna use that instead.

drafts/use-cases/3.obtaining-events.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Story
+As an application user
+I want to display all my events
+So I can find out what awaits me in the nearest future.
```
lanthaler: I would also keep this generic

As an API consumer
I want to be able to retrieve all members of a collection
So that I can process them further.

drafts/use-cases/3.obtaining-events.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Usage
+```javascript
+var client = new HydraClient();
+var data = client.get("/api/events");
```
lanthaler: I would drop the 

here and perhaps introduce a separate page before this that describes following links.

drafts/use-cases/3.obtaining-events.md
```diff
@@ -0,0 +1,110 @@
...
+It knows only that a collection of items will be returned, but the type of these items is not disclosed.
+Also there is no hint before fetching whether the collection is complete or partial.
+Hydra Core Vocabulary allows partial collections, 
+but it's unclear how to denote that fact before fetching the collection.
+This may lead to situation when an application can be "surprised"
+either by receiving a partial collection forcing it to make additional calls to fill the view 
+(in our case a monthly view of the calendar), or by a complete collection which may contain dozens of items 
+(which are not needed at the time being).
+
+#### Hypermedia controls and data separation
```
lanthaler: This is an interesting topic (that will likely yield to lively discussions) but I would prefer this to be moved to a separate doc.

drafts/use-cases/3.obtaining-events.md
```diff
@@ -0,0 +1,110 @@
...
+```javascript
+var client = new HydraClient();
+var data = client.get("/api/events");
+for (var member of data.members) {
+    // do something with the _member_, i.e. display it
+}
+```
+
+### Details
+My application used a discovered either from the entry-point or the API documentation URL to my events.
```
tpluscode: I don't understand this sentence. Some words got lost maybe? :)

drafts/use-cases/3.obtaining-events.md
```diff
@@ -0,0 +1,110 @@
...
+        }
+    ]
+}
+```
+
+It is possible now to display one event in the calendar.
+
+### Considerations
+
+#### Some issues
```
tpluscode: @RubenVerborgh thoughts on the issues below?

My gut tells me that knowing in advance isn't really that important. We may need that information in actual (Partial)Collection's representation. Is it already specified how we'd do that?

Similarly with whether the collection is partial or not should not be of utmost importance. 

In any case let's reflect back on https://www.w3.org/community/hydra/wiki/Pagination#PartialCollectionView and https://www.w3.org/community/hydra/wiki/Collection_Interaction_Flow for earlier ideas and maybe move that to a separate document for discussion.

drafts/use-cases/3.obtaining-events.md
```diff
@@ -0,0 +1,110 @@
...
+Unfortunately, my application is not fully aware of what will come.
+It knows only that a collection of items will be returned, but the type of these items is not disclosed.
+Also there is no hint before fetching whether the collection is complete or partial.
```
RubenVerborgh: That does not seem a problem; in fact, a library should hide this fact from its users and just give access to an asynchronous stream of things.

drafts/use-cases/3.obtaining-events.md
```diff
@@ -0,0 +1,110 @@
...
+Hydra Core Vocabulary allows partial collections, 
+but it's unclear how to denote that fact before fetching the collection.
+This may lead to situation when an application can be "surprised"
+either by receiving a partial collection forcing it to make additional calls to fill the view
```
RubenVerborgh: Applications should always be prepared for this; in fact, they will not see it because the library will hide this. (Successfully implemented this pattern with the LDF query client; results are just streams—the same whether there's 1 or 5,000 pages.)

tpluscode: Stream is a good analogy but detached from an application user. Endless scrolling list however is a user-facing manifestation of the same. The UI can simply append members as long as there is a `next` link

RubenVerborgh: Agree conceptually, but would build the abstraction differently: UI paging does not need to be coupled to API paging. So for me, the UI would depend on a pull-based stream (iterator if you like), which in turn is sourced from the API (paged or not).

drafts/use-cases/3.obtaining-events.md
```diff
@@ -0,0 +1,110 @@
...
+
+```javascript
+{
+    "@graph": [
+        {
+            "@id": "meta:data",
+            "@graph": [
+                {
+                    "@id": "/api/events",
+                    "hydra:operation: [
```
asbjornu: Missing closing quote `"`.

drafts/use-cases/3.obtaining-events.md
asbjornu: I can't quite parse this sentence. 😕

drafts/use-cases/3.obtaining-events.md
asbjornu: I agree on both points: Interesting discussion and something we should discuss elsewhere.

drafts/use-cases/3.obtaining-events.md
```diff
@@ -0,0 +1,110 @@
...
+                    "startDate": "2017-04-19",
+                    "endDate": "2017-04-19"
+                }
+            ]
+        }
+    ]
+}
+```
+
+Alternatives would involve some mixed content type payloads or linking.
```
asbjornu: I think linking is a strong contender, especially with HTTP/2 which would allow the server to push the linked resources to the client, thus reducing the request to a linked resource to a simple cache lookup.

drafts/use-cases/3.obtaining-events.md
alien-mcl: Yeah - copy-paste issue.

drafts/use-cases/3.obtaining-events.md
alien-mcl: Oh, I'd insist on a specific story. I believe that being generic in our current approach led us to nowhere.

drafts/use-cases/3.obtaining-events.md
alien-mcl: I'd stay consistent across the examples. Let's decide whether w shall remove that line everywhere or leave it as it is now.

drafts/use-cases/3.obtaining-events.md
alien-mcl: OK - I'll move those to separate document.

drafts/use-cases/3.obtaining-events.md
lanthaler: I'm fine either way. If you decide to leave it, it somehow needs to be made clear though how that relative URL is transformed to an absolute URL. Maybe we could change it to

    var client = new HydraClient("http://example.com/");

drafts/use-cases/3.obtaining-events.md
lanthaler: I find that out of place as Hydra is not about events. What about having both?

> ### Story 
> As an API consumer
> I want to be able to retrieve all members of a collection
> So that I can process them further.
> 
> ### Example
> As an application user 
> I want to display all my events 
> So I can find out what awaits me in the nearest future.

drafts/use-cases/3.obtaining-events.md
alien-mcl: I'd agree with @RubenVerborgh only when I'd be able to explicitly choose what my app want from the API, leading us to the topic of filtering and projection. I'll leave it from here.

PR #116, opened 2017-04-22
Add use case: Retrieve resource

alien-mcl:
I decided to proceed with more use cases, thus please consider these as a work-in-progres.

I've noticed several editorial-like comments (ugh - need some sleep, I think :)), but I believe we'd like to proceed ASAP, so the sooner more cases we discuss, the better.

I'll try to take all the comments into account _in the meanwhile_.

### DIFF
--- a/drafts/use-cases/4.obtaining-single-event.md
+++ b/drafts/use-cases/4.obtaining-single-event.md
@@ -1,4 +1,4 @@
-## Entry point
+## Obtaining a single event
 
 ### Story
 As an application user
@@ -13,9 +13,9 @@ var event = client.get("/api/events/1");
 ```
 
 ### Details
-After displaying my events in a calendar, I'd like to browse details of a selected event.
-I don't expect calendar will show all attributes, thus I'll have to show those outstanding ones
-in a separate view (i.e. popup, new screen, you name it).
+After displaying all the events in a calendar, user should be able to browse details of a selected event.
+It wouldn't be possible for the calendar to show all of the event's attributes,
+thus there is a need to show those outstanding ones in a separate view (i.e. popup, new screen, etc.).
 The application sends a GET request to the selected event's resource identifier:
 
 ```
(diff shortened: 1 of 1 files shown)


drafts/use-cases/4.obtaining-single-event.md
```diff
@@ -0,0 +1,59 @@
...
+
+```
+GET /api/events/1
+
+HTTP 200 OK
+```
+```javascript
+{
+    "@id": "/api/events/1",
+    "eventName": "Event 1",
```
lanthaler: I would explicitly add the `@type` everywhere.

drafts/use-cases/4.obtaining-single-event.md
```diff
@@ -0,0 +1,59 @@
...
+    "startDate": "2017-04-19",
+    "endDate": "2017-04-19"
+}
+```
+
+Now the user can see what a given event was about.
+
+### Considerations
+
+#### Expansion
```
lanthaler: We may need to find another name for this as one of JSON-LD's main operations is expansion. "Embedding" may be an alternative

drafts/use-cases/4.obtaining-single-event.md
```diff
@@ -0,0 +1,59 @@
+## Obtaining a single event
+
+### Story
+As an application user
```
lanthaler: As discussed, please keep both the abstract story plus the concrete example. Thanks

drafts/use-cases/4.obtaining-single-event.md
alien-mcl: Embedding seems also sounds weird in this context. Maybe _unfolding_?

lanthaler: What about "Partial vs. full representations"?

elf-pavlik: JSON-LD spec uses term *embedding* for that case https://www.w3.org/TR/json-ld/#embedding

alien-mcl: This would make an impression Hydra is for JSON-LD, while same action can be taken in other formats, for which embedding might not exist.

PR #117, opened 2017-04-22
Add use case: Create resource

alien-mcl:
Yet another scenario I acknowledged during my adventures with Hydra.

### DIFF
--- a/drafts/use-cases/5.creating-new-event.md
+++ b/drafts/use-cases/5.creating-new-event.md
@@ -65,7 +65,10 @@ HTTP by referring to HTTP POST method. This has also another weaknes - server ma
 allow to create resources using HTTP PUT, or it may go event further - it can forbid
 the POST method in favour of PUT.
 Specification could be protocol agnostic and give some hints on the logical functions 
-of the operations.
+of the operations. Possible solution could be to use i.e. schema.org actions, but 
+without specification being clear about that, using schema.org actions is just one of 
+possible solutions. There is nothing against using other vocabularies causing the HyDrA  
+Core Vocabylary to be prone to dialects and variations.
 This leads to another weakneses - first thoughs can go in the CRUD direction, but these 
 are not the only approaches available in the wilderness (not to mention that there are 
 voices that put using CRUD in ReST API to question).
--- a/drafts/use-cases/5.creating-new-event.md
+++ b/drafts/use-cases/5.creating-new-event.md
@@ -5,9 +5,14 @@ As an application user
 I want to be able to create new events manually
 So I can have event's I'm aware of added to my calendar.
 
+As an API consumer
+I want to be able to create new resources of a given type
+So I can produce content.
+
 ### Usage
 ```javascript
 var event = {
+    "@type": "schema:Event",
     "eventName": "My brand new event",
     "eventDescription": "Hope it will work",
     "startDate": "2017-04-19",
@@ -17,14 +22,13 @@ var client = new HydraClient();
 var operation = client.get("http://example.com")
     .getApiDocumentation()
     .getEntryPoint()
-    .operations
-    .filter(op => op.method == "POST" && expects == "Event")[0];
+    .getOperations()
+    .filter(op => op.method == "POST" && op.expects == "Event")[0];
 client.invoke(operation, event);
 ```
 
 ### Details
-Browsing events is a must, but creating a new resource is a brand new story.
-First, the application have to be aware of how it can create a new resource.
+The application has to be aware of how it can create a new resource.
 It could get that information either from the API documentation or entry point
 or last response's hypermedia controls. It also have to understand the type 
 of resource in scope (unless it is provided from outside of the application, 
@@ -37,6 +41,7 @@ POST /api/events
 ```
 ```javascript
 {
+    "@type": "schema:Event",
     "eventName": "My brand new event",
     "eventDescription": "Hope it will work",
     "startDate": "2017-04-19",
@@ -55,17 +60,17 @@ Brand new event should be now added to the calendar.
 
 #### Protocol agnostic specification
 While the most popular protocol in use is currently HTTP, ReST API can be implemented 
-on top of other protocols. The example above has a weakness of being strictly bound to 
+on top of other protocols. The example above has a weakneses of being strictly bound to 
 HTTP by referring to HTTP POST method. This has also another weaknes - server may also 
 allow to create resources using HTTP PUT, or it may go event further - it can forbid
 the POST method in favour of PUT.
 Specification could be protocol agnostic and give some hints on the logical functions 
 of the operations.
-This leads to another weakness - first thoughs can go in the CRUD direction, but these 
+This leads to another weakneses - first thoughs can go in the CRUD direction, but these 
 are not the only approaches available in the wilderness (not to mention that there are 
 voices that put using CRUD in ReST API to question).
 
-#### RIA editor form
+#### Creating user interfaces
 It is common that rich UI applications allows their users to create/edit resources
 via some forms full of inputs, lists etc. The question is - how these should be created.
 
@@ -78,7 +83,7 @@ Another, slightly better solution would be to embed some of that logic within th
 hypermedia controls. This is what happens when an HTML page has JavaScripts, CSS's and 
 HTML forms embedded - client receives extra stuff to be used to show the form so it 
 sends the resource back to the server. In our case though, it somehow puts on to our 
-API a burdain of being aware of the client UI capabilities, which seems to be a dead end.
+API a burden of being aware of the client UI capabilities, which seems to be a dead end.
 
 Ultimate solution would be to give the client all the details possible about the resource 
 itself, it's structure, data types etc., so the client can use these on it's own.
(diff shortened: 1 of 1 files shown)


drafts/use-cases/5.creating-new-event.md
```diff
@@ -0,0 +1,88 @@
+## Creating a new event
+
+### Story
```
lanthaler: As discussed, please keep both the abstract story plus the concrete example. Thanks

drafts/use-cases/5.creating-new-event.md
```diff
@@ -0,0 +1,88 @@
...
+As an application user
+I want to be able to create new events manually
+So I can have event's I'm aware of added to my calendar.
+
+### Usage
+```javascript
+var event = {
+    "eventName": "My brand new event",
```
lanthaler: Please add the `@type` (also in the example below). We shouldn't rely on inference as discussed in a previous pull request

drafts/use-cases/5.creating-new-event.md
```diff
@@ -0,0 +1,88 @@
...
+    "eventDescription": "Hope it will work",
+    "startDate": "2017-04-19",
+    "endDate": "2017-04-19"
+};
+var client = new HydraClient();
+var operation = client.get("http://example.com")
+    .getApiDocumentation()
+    .getEntryPoint()
+    .operations
```
lanthaler: .getOperations() would seem more consistent with the rest of the interface

drafts/use-cases/5.creating-new-event.md
```diff
@@ -0,0 +1,88 @@
...
+    .filter(op => op.method == "POST" && expects == "Event")[0];
+client.invoke(operation, event);
+```
+
+### Details
+Browsing events is a must, but creating a new resource is a brand new story.
+First, the application have to be aware of how it can create a new resource.
```
lanthaler: s/have/has/

drafts/use-cases/5.creating-new-event.md
```diff
@@ -0,0 +1,88 @@
...
+var operation = client.get("http://example.com")
+    .getApiDocumentation()
+    .getEntryPoint()
+    .operations
+    .filter(op => op.method == "POST" && expects == "Event")[0];
+client.invoke(operation, event);
+```
+
+### Details
+Browsing events is a must, but creating a new resource is a brand new story.
```
lanthaler: I'd stop this first sentence as it doesn't add anything.

drafts/use-cases/5.creating-new-event.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+Brand new event should be now added to the calendar.
+
+### Considerations
+
+#### Protocol agnostic specification
+While the most popular protocol in use is currently HTTP, ReST API can be implemented 
+on top of other protocols. The example above has a weakness of being strictly bound to 
+HTTP by referring to HTTP POST method. This has also another weaknes - server may also
```
lanthaler: s/weaknes/weakness/

drafts/use-cases/5.creating-new-event.md
```diff
@@ -0,0 +1,88 @@
...
+HTTP 201 Created
+Location: /api/events/2
+```
+
+Brand new event should be now added to the calendar.
+
+### Considerations
+
+#### Protocol agnostic specification
+While the most popular protocol in use is currently HTTP, ReST API can be implemented
```
lanthaler: I think this section would benefit from being cut in half (or even less). Also, Hydra operations can be typed, so there's no need to look for a POST operation. Indeed it is an anti-pattern.

drafts/use-cases/5.creating-new-event.md
```diff
@@ -0,0 +1,88 @@
...
+The easiest approach is to hard-code these in the application. Well, if the application 
+knows how to handle resources of type *Event*, it would have some of the logic 
+and views already prepared. The issue here is that it somehow puts the whole idea of 
+meta-data to question. If we have tools to fully drive the client, why not use them?
+
+Another, slightly better solution would be to embed some of that logic within the 
+hypermedia controls. This is what happens when an HTML page has JavaScripts, CSS's and 
+HTML forms embedded - client receives extra stuff to be used to show the form so it 
+sends the resource back to the server. In our case though, it somehow puts on to our 
+API a burdain of being aware of the client UI capabilities, which seems to be a dead end.
```
lanthaler: s/burdain/burden/

drafts/use-cases/5.creating-new-event.md
```diff
@@ -0,0 +1,88 @@
...
+HTTP by referring to HTTP POST method. This has also another weaknes - server may also 
+allow to create resources using HTTP PUT, or it may go event further - it can forbid
+the POST method in favour of PUT.
+Specification could be protocol agnostic and give some hints on the logical functions 
+of the operations.
+This leads to another weakness - first thoughs can go in the CRUD direction, but these 
+are not the only approaches available in the wilderness (not to mention that there are 
+voices that put using CRUD in ReST API to question).
+
+#### RIA editor form
```
lanthaler: What does RIA stand for? Rich Internet Application? Please don't use acronyms without introducing them first. In any case, a more descriptive title may work better here. Maybe "Creation of user interfaces" might be slightly better (I can't think of a good title right now)?

drafts/use-cases/5.creating-new-event.md
elf-pavlik: Looking at JSON-LD payload of the entry point `operations` comes as property of that object https://github.com/HydraCG/Specifications/blob/master/drafts/use-cases/1.entry-point.md#details

I guess accessing it directly assumes that client compacts the response with particular JSON-LD context before returning it from `getEntryPoint()`

I'll talk with @bergos about prototyping reference implementation of client in JS. IMO it would help to use snippets which one can actually execute in something like http://jsbin.com

alien-mcl: This touches a topic of the internal client's data model. During my exercises with implementing a hydra client I had to use completely JSON-LD detached model.
The examples I'm using are not directly bound to the JSON-LD payloads as well. I can easily imagine _operations_ to be a property *getter*. It can also be a framed JSON-LD.

elf-pavlik: I'd propose to start with minimal expectations toward all the pseudo-code in use cases and update it later based on actual reference implementations of hydra client.

lanthaler: That would be fantastic @elf-pavlik

drafts/use-cases/5.creating-new-event.md
alien-mcl: I remember that in early Hydra spec there were specialized operations related to deleting or creating resources, but I think these were abandoned. Without some features available out of box this anti-pattern may be the easiest to implement.
Typed operations requires some extra implementation effort on the server side and also enforces client to understand these custom typed operations. I'm not sure it is a good direction.

drafts/use-cases/5.creating-new-event.md
alien-mcl: Ok  - I'll go with "Creating user interfaces".

drafts/use-cases/5.creating-new-event.md
tpluscode: What a coincidence that I just wrote about custom operation types too.

Why yo you think it's not a good direction? Typed operations can bear additional semantics which should aid in creating the UI and driving interaction. Not unlike link relations. Possibly with both types explicit for visibility:

``` json
{
  "@type": [ "hydra:Operation", "api:registerUser" ]
}
```

alien-mcl: I'm just saying that some things should be available out of box so consumers won't fall into an anty-pattern in haste. I'm also worried that with no other options, typing operations will result in dozens of dialect-specific Hydra driven applications incapable of talking to each other.
I believe this whole work with use cases is just about that - what can be done with the spec and haw far we can go with it without customization.
Each customization indeed should not be blocked, but neither should be enforced.

elf-pavlik: I appreciate that this section mentions creating new resources with PUT (which I currently explore in some prototyping code). I hope it will stay as part of this use case.
IMO `.filter(op => op.method == "POST" && op.expects == "Event")[0];` shouldn't block this PR and instead create an issue to capture it and patiently discuss it there. Once we find resolution we can update this UC with new PR.

elf-pavlik: BTW you could use `find` instead of `filter` if you just want a single value returned `.find(op => op.method == "POST" && op.expects == "Event")`

lanthaler: We said we would like to use Schema.org actions for this. So you could type it as `hydra:Operation` and `http://schema.org/CreateAction` similar to what @tpluscode proposed.

elf-pavlik: or https://schema.org/AddAction
> The act of editing by adding an object to a collection.

Actually in this case we seem to have combination of *Create* & *Add* as one action :)
Just *add* could get used when adding already existing photo (denoted by IRI) to an album represented as collection.

Issue #25, opened 2014-01-30
Plural properties should become singular


lanthaler:
**_[Raised by Ruben Verborgh](http://lists.w3.org/Archives/Public/public-hydra/2014Jan/0045.html):**_

Dear all,

TL;DR: Since the range of the following properties is not a list, I think they should be singular:
- hydra:mappings
- hydra:members
- hydra:operations
- hydra:statusCodes
- hydra:supportedClasses
- hydra:supportedOperations
- hydra:supportedProperties

In the RDF data model, a property creates a relation between a subject and an object.
For instance, :subject :hasP :object means that “subject has object as P".
We see examples of that in all popular vocabularies. Examples:
RDF – :me rdf:type foaf:Person, foaf:Agent.
FOAF – :me foaf:interest, :Web, :Hypermedia.
Dublin Core – ;document dc:hasPart :part1, :part2.

As you can see, it doesn't matter whether one or multiple properties are expected:
multiple objects can be attached to the same subject; but all are different relations.
The property is singular because it describes a relation between two (singular) things.
The only time we would encounter a plural, is if the object is a list/collection.

If we look at Hydra, this is not defined as one would expect:

```
{
  "@id": "hydra:mappings",
  "@type": "rdf:Property",
  "comment": "The variable-to-property mappings of the IRI template.",
  "domain": "hydra:IriTemplate",
  "label": "mappings",
  "range": "hydra:IriTemplateMapping",
  "status": "testing"
},
```

This can lead to situations such as

```
:Template hydra:template "/issues{?q}" ;
    hydra:mappings :q.
:q a hydra:IriTemplateMapping;
    hydra:property hydra:freetextQuery;
    hydra:required true;
    hydra:variable "q" .
```

So the “mappings” of /issues{?q} is actually a single mapping.
Or in other words: /issues{?q} has "mappings” :q,
but :q is only a single mapping, not a collection of mappings.
Let me trim down the example to show this contradiction better:

```
:Template hydra:mappings :q.
:q a hydra:IriTemplateMapping.
```

Either :q should be hydra:IriTemplateMappings,
or the property should be singular: "hydra:mapping".

The case of “statusCodes” shows to how much confusion this leads:
we have both a “statusCode” and a ”statusCodes” property.
The situation would be that a resource can lead to “statusCodes" X, Y, Z,
each of which have a “statusCode” 200, 400, 500, respectively.
I think the correct way is that “statusCodes” becomes “status”,
[...]

PR #120, opened 2017-05-01
"@type": "@id" in JSON-LD context of use cases

elf-pavlik:
as reported in https://github.com/HydraCG/Specifications/pull/113#pullrequestreview-35602303

https://www.w3.org/TR/json-ld/#h_note_1
>  It is worth noting that the `@type` keyword is not only used to specify the type of a node but also to express typed values (as described in section 6.4 Typed Values) and to type coerce values (as described in section 6.5 Type Coercion). **Specifically, `@type` cannot be used in a context to define a node's type**. For a detailed description of the differences, please refer to section 6.4 Typed Values.

### DIFF
--- a/drafts/use-cases/0.intro.md
+++ b/drafts/use-cases/0.intro.md
@@ -46,7 +46,7 @@ These would be:
             "@id": "hydra:title",
             "@type": "xsd:string"
         },
-        "operations": {
+        "operation": {
             "@id": "hydra:operation",
             "@type": "@id"
         },
--- a/drafts/use-cases/0.intro.md
+++ b/drafts/use-cases/0.intro.md
@@ -56,25 +56,25 @@ These would be:
         },
         "returns": {
             "@id": "hydra:returns",
-            "@type": "@id"
+            "@type": "@vocab"
         },
         "expects": {
             "@id": "hydra:expects",
-            "@type": "@id"
+            "@type": "@vocab"
         },
-        "supportedClasses": {
+        "supportedClass": {
             "@id": "hydra:supportedClass",
-            "@type": "@id"
+            "@type": "@vocab"
         },
-        "supportedProperties": {
+        "supportedProperty": {
             "@id": "hydra:supportedProperty",
-            "@type": "@id"
+            "@type": "@vocab"
         },
         "property": {
             "@id": "hydra:property",
-            "@type": "@id"
+            "@type": "@vocab"
         },
-        "members": {
+        "member": {
             "@id": "hydra:member",
             "@type": "@id"
         },
--- a/drafts/use-cases/0.intro.md
+++ b/drafts/use-cases/0.intro.md
@@ -33,7 +33,7 @@ These would be:
 - to simplify, the host was dropped from all Urls. We can assume an example like http://example.com/.
 
 ### JSON-LD context
-```javascript
+```json
 {
     "@context": {
         "hydra": "http://www.w3.org/ns/hydra/core#",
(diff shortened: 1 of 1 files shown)


alien-mcl:
Hmm. I'm not a JSON-LD expert, but what about http://json-ld.org/spec/latest/json-ld/#dfn-coercion ?
There is an example using same approach. The difference is the example uses an xsd:integer, which is not a class but a datatype, but this shouldn't matter. I have no issues with fixing the issues, but I'd like to be 100% sure that it is indeed an issue.

tpluscode:
That's right. Coercion.

For a JSON object `{ "key": "value" }`

`"@type": "URI"` means that the string value is a typed literal. Always.
`"@typr": "@id"` means that the string value is an identifier.

drafts/use-cases/0.intro.md
```diff
@@ -48,31 +48,31 @@ These would be:
...
+            "@type": "@id"
         },
         "method": {
             "@id": "hydra:method",
             "@type": "xsd:string"
         },
         "returns": {
             "@id": "hydra:returns",
-            "@type": "hydra:Class"
+            "@type": "@id"
```
lanthaler: This, `expects`, `supportedClass`, and `property` should probably be `"@type": "@vocab"`.
I think in general, we should have the same definitions as in the official context that can be found at https://github.com/HydraCG/Specifications/blob/master/spec/latest/core/core.jsonld

drafts/use-cases/0.intro.md
```diff
@@ -48,31 +48,31 @@ These would be:
...
         },
         "expects": {
             "@id": "hydra:expects",
-            "@type": "hydra:Class"
+            "@type": "@id"
         },
         "supportedClasses": {
```
lanthaler: Please change this to `supportedClass` (singular) to just be a shortcut and not an alias

drafts/use-cases/0.intro.md
```diff
@@ -48,31 +48,31 @@ These would be:
...
             "@id": "hydra:supportedClass",
-            "@type": "hydra:Class"
+            "@type": "@id"
         },
         "supportedProperties": {
```
lanthaler: Same here. Please use `supportedProperty` instead

drafts/use-cases/0.intro.md
```diff
@@ -48,31 +48,31 @@ These would be:
...
             "@id": "hydra:supportedProperty",
             "@type": "@id"
         },
         "property": {
             "@id": "hydra:property",
-            "@type": "rdf:Property"
+            "@type": "@id"
         },
         "members": {
```
lanthaler: ... and here. This was discussed in #25 and implemented in PR #34.

Thanks

elf-pavlik:
@lanthaler I think I addressed most of your feedback

> The difference is the example uses an xsd:integer, which is not a class but a datatype, but this shouldn't matter.

@alien-mcl it does matter, with datatype one gets a Literal and not an IRI (NamedNode). I think we can trust @lanthaler  with JSON-LD nuances since he co-edited&authored that spec.

drafts/use-cases/0.intro.md
```diff
@@ -48,33 +48,33 @@ These would be:
         },
         "operations": {
```
lanthaler: Please make this singular `operation`

lanthaler:
Anyone opposed to merging this change? If I don't hear any objection, I'll merge it in the next days

lanthaler:
Thanks for these fixes @elf-pavlik

PR #122, opened 2017-05-03
Add use case: Update resource

alien-mcl:
Yet another use case, when an imaginatory Hydra client updates an event.

### DIFF
--- a/drafts/use-cases/6.updating-event.md
+++ b/drafts/use-cases/6.updating-event.md
@@ -14,7 +14,9 @@ So I can amend content.
 var client = new HydraClient();
 var event = client.get("/api/events/1")
 event.eventName = "Some new event name.";
-client.invoke(event.getOperations().filter(op => op.method == "PUT")[0], event);
+client.invoke(
+    event.getOperations()
+        .filter(op => op[@type].find(type => type === schema.ReplaceAction)[0], event);
 ```
 
 ### Details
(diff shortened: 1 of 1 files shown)


drafts/use-cases/6.updating-event.md
```diff
@@ -0,0 +1,80 @@
...
+As an API consumer
+I want to be able to update existing resources of a given type
+So I can amend content.
+
+### Usage
+```javascript
+var client = new HydraClient();
+var event = client.get("/api/events/1")
+event.eventName = "Some new event name.";
+client.invoke(event.getOperations().filter(op => op.method == "PUT")[0], event);
```
elf-pavlik: Related to discussed in https://github.com/HydraCG/Specifications/pull/117#discussion_r114505359

potentially http://schema.org/ReplaceAction could get used

drafts/use-cases/6.updating-event.md
```diff
@@ -0,0 +1,80 @@
...
+```
+The event should be now updated.
+
+### Considerations
+
+#### What does the term "update" means?
+As mentioned in the [previous use case](/4.obtaining-single-event.md), application is not aware 
+of what kind of data will be provided by the server, thus a simple question may be asked:
+
+*What gets updated?*
```
elf-pavlik: I think we could consider some use cases that use HTTP PATCH, maybe worth mentioning it here?

alien-mcl: Well, PATCH is strictly HTTP related. It also does not define how the changes should be described. Only reasonable and bullet-proof solution I can see now would be a list of triples to be retracted and triples to be added.

elf-pavlik: Solid uses `application/sparql-update` with DELETE and INSERT statements https://github.com/solid/solid-spec/blob/master/api-rest.md#alternative-using-sparql-1

lanthaler: We can certainly leave a TODO or something but I wouldn't invest much time in PATCH for now. It really opens whole new can of worms.

PR #123, opened 2017-05-07
Add use case: Search resources

alien-mcl:
The force was strong in me this day - there will be two cases related to search and filtering in general.


drafts/use-cases/7.searching-events.md
```diff
@@ -0,0 +1,120 @@
+﻿## Searching for events
+
+### Story
+As an application user
+I want to be able to quickly search through my events
+So I find events I'm interested in.
+
+As an API consumer
+I want to be able to search through resources
```
lanthaler: Should we make this "through a collection of resources"? I don't think we have a good story about search in general yet

drafts/use-cases/7.searching-events.md
```diff
@@ -0,0 +1,120 @@
...
+So I can find those interesting to me.
+
+### Usage
+```javascript
+var client = new HydraClient();
+var data = client.get("/api/events?search=some text");
```
lanthaler: As discussed on the call, let's elaborate this a bit. It would be great for a client to be able to discover the search affordance, enumerate and set the available parameters and then execute the request

drafts/use-cases/7.searching-events.md
```diff
@@ -0,0 +1,120 @@
...
+for (var member of data.members) {
+    // do something with the _member_, i.e. display it
+}
+```
+
+### Details
+The application should be able to search through events,
+so the user can narrow number and importance of the events displayed.
```
lanthaler: narrow number -> narrow down the

drafts/use-cases/7.searching-events.md
```diff
@@ -0,0 +1,120 @@
...
+Communication would look like this:
+
+```
+GET /api/events?search=search%20phrase
+
+HTTP 200 OK
+```
+```json
+{
+    "@id": "/api/events?search=search phrase",
```
lanthaler: Did you intentionally replace the `%20` in the URL here with a space?

drafts/use-cases/7.searching-events.md
```diff
@@ -0,0 +1,120 @@
...
+    "members": [
+        {
+            "@id": "/api/events/1",
+            "eventName": "Event 1",
+            "eventDescription": "Some event 1",
+            "startDate": "2017-04-19",
+            "endDate": "2017-04-19"
+        }
+    ],
+    "search": {
```
lanthaler: This is exactly how it was intended 👍

drafts/use-cases/7.searching-events.md
```diff
@@ -0,0 +1,120 @@
...
+```turtle
+</api/events> hydra:search </api-doc/events#search-iri-template> .
+```
+
+Above's example of a triple with a templated link
+seems inconvinient as in order to define a custom 
+templated link, one should create a separate resource.
+IRI of that resource shall be nicely foldable to QIri form
+in order to improve readability, resulting in a number 
+of fake namespaces.
```
lanthaler: Is this trying to say that you would like to have a generic predicate for template links? How would a client understand which one to pick then?

drafts/use-cases/7.searching-events.md
```diff
@@ -0,0 +1,120 @@
...
+
+Also a location of the template itself within the graph is not clear. 
+The example above puts it as a related resource of the 
+*hydra:search* predicate, but only to fill the gap as
```
lanthaler: It is a related resource of `/api/events`. The relation type is `hydra:search`. So the statement is saying that it is possible to search `/api/events` by using the IRI template that `hydra:search` points to

drafts/use-cases/7.searching-events.md
alien-mcl: OK. I'll have it changed,

drafts/use-cases/7.searching-events.md
alien-mcl: Well, didn't think of it. For the sake of readability I'd leave it as is.

drafts/use-cases/7.searching-events.md
alien-mcl: The whole concept was about having custom templated predicates. Client could find them out i.e. by checking out the type of a resource. The thing is that these custom predicates could be automatically generated by servers, thus the urls created could be less readable than those created by i.e. programmers.

drafts/use-cases/7.searching-events.md
alien-mcl: Indeed - but it is expressed with a PagedCollectionView instance that could hold that _search_ predicate (at least this is the only hint in the specification).

drafts/use-cases/7.searching-events.md
alien-mcl: I'll try to add few lines to address that

Issue #118, opened 2017-04-24
Specifying operations on TemplatedLink objects


tpluscode:
I wasn't sure how to phrase the title but I hope it's clear enough. 

I'd asked in the past about a away to specify operations which have a variable target (ie. an `IriTemplate`). It just dawned on me that I may have been looking at this from the wrong side. Please help me verify my assumptions and possible usage.

Given that the templated link is a property, and a `hydra:Resource`, it should be possible to attach an operation to it, right? So, assuming `ex:tag` is a `hydra:TemplatedLink`, would it be legal to have:

``` json
{
  "ex:tag": {
    "@type": "IriTemplate",
    "template": "http://example.com/tag/{tag}",
    "mappings": [{ "variable": "tag" }],
    "operation": [
      {
        "method": "POST",
        "expects": { "@type": "ex:NotificationRequest" }
      },
      {
        "method": "DELETE"
      }
    ]
  }
}
```

I don't see anything that contradicts the semantics of current vocab. Does the above come close to resolving #100.

By extension, I believe that if we allowed hydra:returns on IriTemplate (#50) then it would be possible to take advantage of `hydra:SupportedOperation` to define actions possible to perform on templates targets.

``` json
{
  "ex:tag": {
    "@type": "IriTemplate",
    "template": "http://example.com/tag/{tag}",
    "mappings": [{ "variable": "tag" }],
    "returns": "ex:Tag"
  }
}
```

Where `ex:Tag` could support the `DELETE`/`POST` operations inlined in the first snippet.

cr3a7ure:
Looking the vocabulary diagram and the spec, `hydra:Link` and `hydra:TemplatedLink` seems to have the same meaning. Although I can't see a single reference to a template that it's name suggests.  
Supposing the above is true, we can attach operations to `hydra:TemplatedLink` the same way as `hydra:Link`.

But while `hydra:IriTemplate` has a well defined mapping `hydra:TemplatedLink` seems needless.
So using either of them (which one seems right though?) should be enough to attach an operation to it, while `returns` is contained inside the operation as usual and not at the outer node. 

I'm pretty new to all this, don't take my word.

tpluscode:
AFAIK `hydra:Link` is just a property. It is a hint for a hydra client which tells that client that it can expect something interesting when dereferencing the linked resource.

The `hydra:TemplatedLink` on the other hand doesn't link to a specific resource but to a whole range of resources whose URIs are constructed by substituting the template variables.

[...]

PR #124, opened 2017-05-07
Add use case: Filter resources

alien-mcl:
As promised, filtering use case.


drafts/use-cases/8.filtering-events.md
```diff
@@ -0,0 +1,72 @@
...
+So I can find events I'm interested in.
+
+As an API consumer
+I want to be able to filter resources
+So I can find those interesting to me.
+
+### Usage
+```javascript
+var client = new HydraClient();
+var data = client.get("/api/events?schema:eventName=.*test.*");
```
lanthaler: Same as on the other PR I would suggest to extend this a bit

drafts/use-cases/8.filtering-events.md
```diff
@@ -0,0 +1,72 @@
...
+
+From technical point of view, this behavior could be achieved
+with special purpose virtual predicates 
+(i.e.:  imaginatory hydra:limit and hydra:offset).
+The argument against is that these are not direct properties 
+of resources being filtered.
+Supporting argument would be that similar approach is used 
+in i.e. MS SQL pagination queries, where the row
+(which can be a one-to-one relation to an entity in our cases)
+is extended with a special purpose ROW_NUMBER value
```
lanthaler: The problem is that not all backends would be able to implement. So we would need to provide some mechanisms to advertise support for it.

asbjornu: Each database engine basically has its own implementation of paging. There's `take`, `limit`, `top`, `skip`, etc. Hydra should be agnostic to how the serialization of this looks like and instead focus on the abstract concept, no?

RubenVerborgh: We might want to postpone this for now; let's not set such difficult paging decisions in stone before we actually have completed paging.

tpluscode: These are connected concepts. Again, it's why I oppose putting too much into the core vocabulary. We need a minimal set of extension point to support any paging/filtering scheme. And if you ask me, the current `PartialCollactionView` comes very close.

alien-mcl: That's why I put it in the considerations. I also agree about not having too much in the core.
Still, currently spec neither have that covered nor allows any extension other than generic IriTemplate (which may end with stumbling upon the issue #118).

PR #125, opened 2017-06-17
use schema.org actions

elf-pavlik:
I also added JSON-LD context to all snippets. Since my editor removed quite a bit trailing whitespaces, better to look at a diff which ignores whitespace changes: https://github.com/HydraCG/Specifications/pull/125/files?w=1

### DIFF
--- a/drafts/use-cases/1.1.security-considerations.md
+++ b/drafts/use-cases/1.1.security-considerations.md
@@ -21,12 +21,6 @@ There are several possibilities here. One of them would be to use a *@graph* not
                 "@type": "hydra:Collection",
                 "operation": [
                     {
-                       "@type": ["hydra:Operation", "schema:ViewAction"],
-                        "title": "List events",
-                        "method": "GET",
-                        "returns": "hydra:Collection"
-                    },
-                    {
                        "@type": ["hydra:Operation", "schema:CreateAction"],
                         "title": "Create new event",
                         "method": "POST",
--- a/drafts/use-cases/1.1.security-considerations.md
+++ b/drafts/use-cases/1.1.security-considerations.md
@@ -10,23 +10,30 @@ There are several possibilities here. One of them would be to use a *@graph* not
 
 ```json
 {
+    "@context": "/api/context.jsonld",
     "@graph": [
         {
             "@id": "/api",
-            "operation": [
-                {
-                    "@type": "hydra:Operation",
-                    "title": "List events",
-                    "method": "GET",
-                    "returns": "hydra:Collection"
-                },
-                {
-                    "@type": "hydra:Operation",
-                    "title": "Create new event",
-                    "method": "POST",
-                    "expects": "schema:Event"
-                }
-            ]
+            "@type": "hydra:EntryPoint",
+            "#events": {
+                "@id": "/api/events",
+                "title": "List of events",
+                "@type": "hydra:Collection",
+                "operation": [
+                    {
+                       "@type": ["hydra:Operation", "schema:ViewAction"],
+                        "title": "List events",
+                        "method": "GET",
+                        "returns": "hydra:Collection"
+                    },
+                    {
+                       "@type": ["hydra:Operation", "schema:CreateAction"],
+                        "title": "Create new event",
+                        "method": "POST",
+                        "expects": "schema:Event"
+                    }
+                ]
+            }
         },
         {
             "rdfs:label": "Some other resource somehow related to the main one."
--- a/drafts/use-cases/1.1.security-considerations.md
+++ b/drafts/use-cases/1.1.security-considerations.md
@@ -15,7 +15,7 @@ There are several possibilities here. One of them would be to use a *@graph* not
         {
             "@id": "/api",
             "@type": "hydra:EntryPoint",
-            "#events": {
+            "examplevocab:events": {
                 "@id": "/api/events",
                 "title": "List of events",
                 "@type": "hydra:Collection",
--- a/drafts/use-cases/1.entry-point.md
+++ b/drafts/use-cases/1.entry-point.md
@@ -28,18 +28,27 @@ HTTP 200 OK
 ```
 ```json
 {
+    "@context": "/api/context.jsonld",
     "@id": "/api",
     "@type": "hydra:EntryPoint",
-    "events": {
-        "@id": "/api/events", 
+    "#events": {
+        "@id": "/api/events",
         "title": "List of events",
         "@type": "hydra:Collection",
-        "operation": {
-            "@type": "hydra:Operation",
-            "title": "Create new event",
-            "method": "POST",
-            "expects": "schema:Event"
-        }
+        "operation": [
+            {
+               "@type": ["hydra:Operation", "schema:ViewAction"],
+                "title": "List events",
+                "method": "GET",
+                "returns": "hydra:Collection"
+            },
+            {
+              "@type": ["hydra:Operation", "schema:CreateAction"],
+              "title": "Create new event",
+              "method": "POST",
+              "expects": "schema:Event"
+            }
+        ]
     }
 }
 ```
--- a/drafts/use-cases/1.entry-point.md
+++ b/drafts/use-cases/1.entry-point.md
@@ -23,9 +23,12 @@ details on what else can be done with that API from that very point.
 
 ```http
 GET /api
+```
 
+```http
 HTTP 200 OK
 ```
+
 ```json
 {
     "@context": "/api/context.jsonld",
@@ -37,16 +40,10 @@ HTTP 200 OK
         "@type": "hydra:Collection",
         "operation": [
             {
-               "@type": ["hydra:Operation", "schema:ViewAction"],
-                "title": "List events",
-                "method": "GET",
-                "returns": "hydra:Collection"
-            },
-            {
-              "@type": ["hydra:Operation", "schema:CreateAction"],
-              "title": "Create new event",
-              "method": "POST",
-              "expects": "schema:Event"
+                "@type": ["hydra:Operation", "schema:CreateAction"],
+                "title": "Create new event",
+                "method": "POST",
+                "expects": "schema:Event"
             }
         ]
     }
--- a/drafts/use-cases/1.entry-point.md
+++ b/drafts/use-cases/1.entry-point.md
@@ -34,7 +34,7 @@ HTTP 200 OK
     "@context": "/api/context.jsonld",
     "@id": "/api",
     "@type": "hydra:EntryPoint",
-    "#events": {
+    "examplevocab:events": {
         "@id": "/api/events",
         "title": "List of events",
         "@type": "hydra:Collection",
--- a/drafts/use-cases/4.obtaining-single-event.md
+++ b/drafts/use-cases/4.obtaining-single-event.md
@@ -24,9 +24,12 @@ The application sends a GET request to the selected event's resource identifier:
 
 ```http
 GET /api/events/1
+```
 
+```http
 HTTP 200 OK
 ```
+
 ```json
 {
     "@context": "/api/context.jsonld",
--- a/drafts/use-cases/4.obtaining-single-event.md
+++ b/drafts/use-cases/4.obtaining-single-event.md
@@ -22,13 +22,14 @@ It wouldn't be possible for the calendar to show all of the event's attributes,
 thus there is a need to show those outstanding ones in a separate view (i.e. popup, new screen, etc.).
 The application sends a GET request to the selected event's resource identifier:
 
-```
+```http
 GET /api/events/1
 
 HTTP 200 OK
 ```
-```javascript
+```json
 {
+    "@context": "/api/context.jsonld",
     "@id": "/api/events/1",
     "@type": "schema:Event",
     "eventName": "Event 1",
@@ -50,15 +51,15 @@ whole resource, properties with literal values only, or some other subset.
 Including (or excluding) those additional details would be extremly useful
 for obtaining related resources, i.e. in case of an event application might
 want to expand attendees (or at least some of their details).
-This way some optimizations can be achieved when a single request obtains 
+This way some optimizations can be achieved when a single request obtains
 details that feed more than a single screen which may improve user experience.
 
 #### Projection
 Another interesting feature would be a possibility of projecting a resource.
 It is a commont situation when an application shows an edit form for
-a given resource - in order to assign a value to the i.e. event's attendee property, some kind of 
+a given resource - in order to assign a value to the i.e. event's attendee property, some kind of
 UI element should be presented (let's say a search-as-you-type like input)
 which would display projected resources - concatenation of the first and last name.
 
-The issue here is that the syntax for such a feature would make the spec 
+The issue here is that the syntax for such a feature would make the spec
 incredibly complex.
--- a/drafts/use-cases/5.creating-new-event.md
+++ b/drafts/use-cases/5.creating-new-event.md
@@ -23,8 +23,8 @@ var client = new HydraClient();
 var operation = client.get("http://example.com")
     .getApiDocumentation()
     .getEntryPoint()
-    .onLink('#events')
-    .getOperationOfType(schema.CreateAction);
+    .getLink('#events')
+    .getOperationOfType('http://schema.org/CreateAction');
 client.invoke(operation, event);
 ```
 
@@ -40,6 +40,7 @@ Anyway, assuming all the requirements are met, a request would look like this:
 ```http
 POST /api/events
 ```
+
 ```json
 {
     "@context": "/api/context.jsonld",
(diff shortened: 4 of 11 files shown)


drafts/use-cases/1.1.security-considerations.md
```diff
@@ -10,23 +10,30 @@ There are several possibilities here. One of them would be to use a *@graph* not
...
-                },
-                {
-                    "@type": "hydra:Operation",
-                    "title": "Create new event",
-                    "method": "POST",
-                    "expects": "schema:Event"
-                }
-            ]
+            "@type": "hydra:EntryPoint",
+            "#events": {
```
elf-pavlik: I just used fragment identifier but maybe we should define it in ApiDocumentation instead?

asbjornu: Yes, I think we should use as little "syntactic sugar" in the format as possible. We should perhaps even avoid `@`.

drafts/use-cases/1.1.security-considerations.md
```diff
@@ -10,23 +10,30 @@ There are several possibilities here. One of them would be to use a *@graph* not
...
+                "@id": "/api/events",
+                "title": "List of events",
+                "@type": "hydra:Collection",
+                "operation": [
+                    {
+                       "@type": ["hydra:Operation", "schema:ViewAction"],
```
asbjornu: Is it possible to make this relationship implied, somehow? I.e. that `schema:ViewAction` is a `hydra:Operation`.

tpluscode: I'd keep this explicit. The best we can do is a weak relation. Strong semantics would mean changing the meaning of schema Actions globally. I know it's just an RDF thing but still.

And regardless, any `subclass` relationship of operations means additional work for the client to perform. When specified explicitly it's a simple matter of looking for `hydra:Operation`

drafts/use-cases/1.1.security-considerations.md
tpluscode: You mean you'd prefer a full URI? And what's wrong `@`? It's a core JSON-LD feature and explicitly marks format-specific parts. I think it's better then aliasing in the `@context` which would make details more obscure

drafts/use-cases/5.creating-new-event.md
```diff
@@ -22,25 +23,26 @@ var client = new HydraClient();
 var operation = client.get("http://example.com")
     .getApiDocumentation()
     .getEntryPoint()
-    .getOperations()
-    .filter(op => op.method == "POST" && op.expects == "Event")[0];
+    .onLink('#events')
```
tpluscode: I'd rename this simply `getLink()` not to suggest some kind of fluent API

drafts/use-cases/1.1.security-considerations.md
elf-pavlik: @asbjornu if you would like to alias '@type' to 'type' in JSON-LD `@context` could you please propose it in dedicated PR?

drafts/use-cases/1.1.security-considerations.md
elf-pavlik: I don't think we should expect reasoning based on *rdfs:subClassOf* from the clients. If Hydra had those few basic operations defined in a core, then we could expect from implementations to treat them as subclass of `hydra:Operation` but they got removed in #11

drafts/use-cases/5.creating-new-event.md
elf-pavlik: I could do that. @alien-mcl when you introduced that naming, did `get` prefix supposed to suggest async function or you didn't intend any naming convention for sync vs. async?

alien-mcl: I'm not sure what _onLink('#events')_ means.
As for your question, it is still a pseudo code, thus I didn't focus on async or non async conventions - it didn't matter here.
I also woudln't use appendices like _Async_in methods i.e. _getApiDocumentationAsync_. Nowadays it is natural to use async methods, thus sync methods would get such an appendix.
I also didn't want to imply any convention for async methods (i.e. promises, async keyword or callbacks) - it didn't matter here as well. In my latest implementation I started I'm using an async keyword.

drafts/use-cases/1.1.security-considerations.md
asbjornu: My line of thought is that anything not native to plain JSON will look alien to most people already familiar with plain JSON. Just the idea of using URIs as opaque identifiers (instead of UUIDs or integers) seem crazy to most people at first.

To keep the number of alien elements to a minimum, it would help to keep the syntax as plain as possible. I'm not decided on whether dropping the `@` prefix is a good idea or not, but it's worth investigating, imho.

drafts/use-cases/1.1.security-considerations.md
asbjornu: Ok, I agree.

drafts/use-cases/1.1.security-considerations.md
alien-mcl: I'm not sure how _@_ could look alien. From JSON point of view it's just another property with a _@_ char at the beginning. Indeed it marks that property as something special. HAL uses special properties with an underscore char *_*.

lanthaler: > I just used fragment identifier but maybe we should define it in ApiDocumentation instead?

I think the right way would be to define it in a separate "my-vocab.jsonld" file. This would then also show how people can define their own terms

drafts/use-cases/1.1.security-considerations.md
```diff
@@ -10,23 +10,30 @@ There are several possibilities here. One of them would be to use a *@graph* not
...
+                        "title": "List events",
+                        "method": "GET",
+                        "returns": "hydra:Collection"
+                    },
+                    {
+                       "@type": ["hydra:Operation", "schema:CreateAction"],
+                        "title": "Create new event",
```
lanthaler: Nit: The indentation is different here

drafts/use-cases/1.1.security-considerations.md
lanthaler: Let's keep them separate and explicit for now. We can always optimize it later if we want to

drafts/use-cases/1.entry-point.md
```diff
@@ -28,18 +28,27 @@ HTTP 200 OK
...
-            "@type": "hydra:Operation",
-            "title": "Create new event",
-            "method": "POST",
-            "expects": "schema:Event"
-        }
+        "operation": [
+            {
+               "@type": ["hydra:Operation", "schema:ViewAction"],
+                "title": "List events",
+                "method": "GET",
```
lanthaler: I think at some point we should decide whether we want people to explicitly document unparametrized GETs or not. I for one think it provides close to zero value.

drafts/use-cases/4.obtaining-single-event.md
```diff
@@ -22,13 +22,14 @@ It wouldn't be possible for the calendar to show all of the event's attributes,
 thus there is a need to show those outstanding ones in a separate view (i.e. popup, new screen, etc.).
 The application sends a GET request to the selected event's resource identifier:
 
-```
+```http
 GET /api/events/1
 
 HTTP 200 OK
```
lanthaler: Using two blocks makes it appear as the HTTP 200 OK is still part of the request. I'd say we use 1 or 3 (request headers, response headers, payload)

drafts/use-cases/6.updating-event.md
```diff
@@ -14,25 +14,25 @@ So I can amend content.
 var client = new HydraClient();
 var event = client.get("/api/events/1")
 event.eventName = "Some new event name.";
-client.invoke(
-    event.getOperations()
-        .filter(op => op[@type].find(type => type === schema.ReplaceAction)[0], event);
+var operation = event.getOperationOfType(schema.ReplaceAction)
```
lanthaler: Is schema.ReplaceAction defined somewhere? If not, should we just use the full URL in these examples instead?

drafts/use-cases/1.entry-point.md
tpluscode: You mean whether a GET should be implicitly supported by a resource?

drafts/use-cases/1.1.security-considerations.md
alien-mcl: Maybe we shall put a statement in the spec about having those explicit? Spec is silent about this kind assumptions leaving lot of room for weirdness in that matter.

drafts/use-cases/1.entry-point.md
alien-mcl: Well, there is nothing that can stop client from using any non-blank resource identifier from an RDF statement to de-reference it i.e. via HTTP GET for http based IRI's - I think it is safe to assume that GET is supported by any resource.

drafts/use-cases/1.1.security-considerations.md
tpluscode: Nah, I dunno. The spec defines `hydra:operation` to have range `hydra:Operation`. That's all you need really. Isn't that an answer to @asbjornu's initial comment?

drafts/use-cases/1.entry-point.md
tpluscode: Yes but everything else in Hydra is **explicit, or try at your own peril**. Ie. if `GET` operation is absent, one could assume that `GET` is not supported, right?

I like this way of everything being explicit because you may have resources which are write-only. How would you model that?

alien-mcl: Donno, operation types? We are already marking some of them with schema actions. Just an idea.
Maybe an opposite predicate, i.e. forbidden method?

drafts/use-cases/1.1.security-considerations.md
alien-mcl: Is there any chance an operation will be unrelated to any resource with 'hydra:operation' or 'hydra:supportedOperation' ? What client should do in that case?

drafts/use-cases/5.creating-new-event.md
elf-pavlik: > I'm not sure what onLink('#events') means.

From EntryPoint client should follow the `#events` *hydra:Link* (again, which definition I could move to JSON-LD context of this API instead of using fragment) which pints to the collection of events and we want to get particular operation on that collection resource.

drafts/use-cases/1.entry-point.md
lanthaler: > Ie. if GET operation is absent, one could assume that GET is not supported, right?

No, Hydra is based on the open world assumption. Unless something is explicitly marked as being forbidden you can't draw such a conclusion.

> you may have resources which are write-only. How would you model that?

Currently that isn't possible. The client would need to figure that out at runtime. Do we have such a use case?

drafts/use-cases/1.1.security-considerations.md
lanthaler: > Nah, I dunno. The spec defines `hydra:operation` to have range `hydra:Operation`. That's all you need really. Isn't that an answer to @asbjornu's initial comment?

Currently it is but we may wanna move away from RDFS' range and use schema.org's rangeInclude instead which doesn't support such inferences. 

> Is there any chance an operation will be unrelated to any resource with 'hydra:operation' or 'hydra:supportedOperation' ? What client should do in that case?

No, currently that's not possible.

alien-mcl: So there shouldn't be any operation disconnected from any relation with resource, link or class? I think spec should mention it somehow as currently there is nothing against pushing a standalone resource of type hydra:Operation which can be used by a client.

drafts/use-cases/1.entry-point.md
alien-mcl: No, we don't. But I could imagine such a resource, i.e. gateway for uploading files

drafts/use-cases/5.creating-new-event.md
alien-mcl: Then I'd agree with @tpluscode about renaming it to `getLink`. I thini `onLink` is somehow related to events.

drafts/use-cases/1.entry-point.md
tpluscode: > Hydra is based on the open world assumption

Well sure, no one stops your from dereferencing any resource. But as the client (library), I'd expect clear hints on what the intended usage is. Otherwise you could defend a stance that any HTTP request is valid and the client would brute-force anything. Especially that Hydra has no way to mark operations as forbidden

drafts/use-cases/1.1.security-considerations.md
tpluscode: I don't recall anything like this and don't know what the use is for such detached operations

drafts/use-cases/5.creating-new-event.md
lanthaler: +1

drafts/use-cases/1.1.security-considerations.md
lanthaler: > So there shouldn't be any operation disconnected from any relation with resource, link or class?

What would be the use case? An operation needs to be connected to a resource (or an IRI template) to be executable.

drafts/use-cases/1.1.security-considerations.md
asbjornu: @alien-mcl: All I can say is how I reacted myself and how I experience others react to the JSON-LD syntax. To programmers, all syntactic sugar comes with a bit of overhead. It needs to be visually parsed and understood, then it takes quite a bit of time getting used to.

Learning the syntax of Ruby and Perl is therefore more taxing than learning the syntax of Lisp, for instance. Note that learning the syntax is different from learning the language. Learning the language (keywords, statements, structure) is not the same as learning its syntax.

drafts/use-cases/1.1.security-considerations.md
asbjornu: As long as it's possible to define operations that span several different resources, I guess it's fine.

drafts/use-cases/1.entry-point.md
asbjornu: I'm with @tpluscode here. The "open world assumption" of RDF is a bit too much to take in and won't go well with most developers. On one side we have undefined operations that may or may not work as discussed here, on the other we have terms in referenced vocabularies that may or may not be present.

Being able to scope the implementation of a client application to what the server **actually has to offer** is imho very important. Saying that the scope is infinite is of course very flexible, but it makes the task of implementing a sensible client application well nigh impossible.

elf-pavlik:
I believe that in last commit I have addressed most of the feedback. I would like to propose to merge this PR as long as no one objects any of the changes it introduces. Everyone can create dedicated PR to propose further improvements...

drafts/use-cases/1.1.security-considerations.md
elf-pavlik: I'd like to leave `#events` in this PR and I already plan another PR which would propose alternative approach to `hydra:Link` and rely just on `schema:Event` used together with [`void:classPartition`](http://vocab.deri.ie/void#classPartition). Which would remove that `#events` all together since IMO it doesn't really address need to discover relevant collection in the dataset exposed by this API.

lanthaler: Please let's get rid of `#events`. It won't work unless `@vocab` is specified in the context. If the vocab is set to schema.org or Hydra, it would make this property part of those vocabularies. Certainly not what you intended. Could we change it to `myvocab:events` or something similar for the time being?

drafts/use-cases/1.entry-point.md
lanthaler: > I'd expect clear hints on what the intended usage 

For GETs, that's what `hydra:Resource` ("The class of dereferenceable resources) and `hydra:Link` (its range is `hydra:Resource`) are for.

drafts/use-cases/1.1.security-considerations.md
elf-pavlik: I defined `examplevocab:events` in JSON-LD context but in client pseudo code I used full IRI since we shouldn't assume the client uses the same prefix map as JSON-LD context used by the API. I hope this gives us enough to merge this PR and in future PRs consider available options for how to get rid of that `examplevocab:events` all together and only rely on `schema:Event` when client needs to discover the collection of all events in the dataset.

elf-pavlik: I just shared an idea on how we could use `void:classPartition` in #126

drafts/use-cases/1.1.security-considerations.md
alien-mcl: What about an operation that is also an IriTemplate or a Link? From the other side - how do you connect an operation which is supported by a class? What would be the Url of such an operation?
Just trying to find a hole :)

tpluscode: I think these are valid points but let's save that discussion for later

lanthaler:
As agreed in today's telecom I'll proceed and merge this PR.

PR #130, opened 2017-07-12
Fix GitHub pages update script

lanthaler:


.travis_update_ghpages.sh
```diff
@@ -1,25 +1,43 @@
...
+git clone --branch=gh-pages https://${GH_TOKEN}@github.com/$REPO_NAME gh-pages
 cd gh-pages
-
-# Don't update if already at the latest version
-if [[ `git log -1 --pretty=%B` == *$TRAVIS_COMMIT* ]]; then exit; fi
+mkdir -p drafts/diagram/
 
 # Update documents
-cp -r $REPO_PATH .
+cp -rf $REPO_PATH/drafts/diagram/* drafts/diagram/
```
RubenVerborgh: Note: the reason I copied everything was to have a generic gh-page update script, which works for all future build steps we might need.

.travis_update_ghpages.sh
```diff
@@ -1,25 +1,43 @@
...
 # Commit and push latest version
-git add .
 git config user.name  "Travis"
-git config user.email "travis@travis-ci.org"
+git config user.email "travis@hydra-cg.com"
+git add -A .
+git status
 git commit -m "Update to $TRAVIS_COMMIT."
-git push -fq origin gh-pages 2>&1 > /dev/null
+git push -q origin gh-pages
```
RubenVerborgh: You might want to silence the output; it can contain sensitive information.

lanthaler: Under what circumstances would that happen?

RubenVerborgh: Unsure about all the errors that could happen here, but basically `origin` uses the secret `GH_TOKEN` variable (which I have set in Travis). Perhaps if there is a connection drop or whatever, the reported error could be "could not push to SECRETTOKEN@github.com" and then all my repositories are screwed ;-)

PR #131, opened 2017-07-19
fix iri template expansion in filter use case

elf-pavlik:
it seems that in the snippet we have incorrect expansion of the IRI template

BTW does `hydra:BasicRepresentation` fits here or we should have something different since we use a regular expression?


lanthaler:
Thanks Pavlik. `hydra:BasicRepresentation` is fine here as we neither need type and language information nor do we need to differentiate between IRIs and literals.

Issue #126, opened 2017-06-22
Simplify discovery of collections that contain resources of a certain type


elf-pavlik:
I don't want to wait for merging of #125 and would like to propose ASAP a possibility of addressing one of the issues we started discussing there. In #125 for client to discover collection of all events we rely on a custom property that relates the entry point with that collection `examplevocab:events`.

As alternative I would propose to use [`void:classPartition`](http://vocab.deri.ie/void#classPartition) so the API EntryPoint would change to (given `void` prefix defined in JSON-LD context)

```json
{
    "@context": "/api/context.jsonld",
    "@id": "/api",
    "@type": "hydra:EntryPoint",
    "void:classPartition": {
        "@id": "/api/events",
        "title": "List of events",
        "@type": "hydra:Collection",
        "void:class": "schema:Event",
        "operation": [
            {
                "@type": ["hydra:Operation", "schema:CreateAction"],
                "title": "Create new event",
                "method": "POST",
                "expects": "schema:Event"
            }
        ]
    }
}
```

And the client pseudo code for creating new event would look something like

```javascript
var event = { ... };
var client = new HydraClient();
var operation = client.get("http://example.com")
    .getApiDocumentation()
    .getEntryPoint()
    .getClassPartitionFor('http://schema.org/Event')
    .getOperationOfType('http://schema.org/CreateAction');
client.invoke(operation, event);
```

asbjornu:
Excuse my severe lack of RDF knowledge, but what is `void:classPartition`? Its description makes me none the wiser and your example leaves me both dizzy and puzzled. 😄

tpluscode:
I'm not convinced either. Hardcore RDF people will immediately notice that this implies that `/api` and `/api/events/` are `void:Dataset`. Is that your intention? I think they aren't because the notion of dataset is somewhat disconnected from an API (set of resources)

elf-pavlik:
> Excuse my severe lack of RDF knowledge, but what is void:classPartition? Its description makes me none the wiser and your example leaves me both dizzy and puzzled.

https://www.w3.org/TR/void/#class-property-partitions

> I'm not convinced either. Hardcore RDF people will immediately notice that this implies that /api and /api/events/ are void:Dataset. Is that your intention? I think they aren't because the notion of dataset is somewhat disconnected from an API (set of resources)

API provides an interface to a dataset, and I see no problem with considering a collection a subset aft that dataset (so also a dataset)

[...]

PR #132, opened 2017-08-13
Use manages block to advertise type of collection members

elf-pavlik:
Thise PR tries to address #126 by reusing already agreed on [Collection Design](https://www.w3.org/community/hydra/wiki/Collection_Design) and not introducing any new terms to hydra vocab.

---
This change is [Reviewable](https://reviewable.io/reviews/hydracg/specifications/132)

### DIFF
--- a/drafts/use-cases/0.intro.md
+++ b/drafts/use-cases/0.intro.md
@@ -42,7 +42,6 @@ These would be:
         "xsd": "http://www.w3.org/2001/XMLSchema#",
         "owl": "http://www.w3.org/2002/07/owl#",
         "schema": "http://schema.org/",
-        "examplevocab": "http://example.com/vocab#",
         "title": {
             "@id": "hydra:title",
             "@type": "xsd:string"
@@ -75,8 +74,24 @@ These would be:
             "@id": "hydra:property",
             "@type": "@vocab"
         },
+        "subject": {
+            "@id": "hydra:subject",
+            "@type": "@id"
+        },
+        "object": {
+            "@id": "hydra:object",
+            "@type": "@id"
+        },
         "member": {
             "@id": "hydra:member",
+            "@type": "@id"
+        },
+        "collection": {
+            "@id": "hydra:collection",
+            "@type": "@id"
+        },
+        "manages": {
+            "@id": "hydra:manages",
             "@type": "@id"
         },
         "totalItems": {
--- a/drafts/use-cases/1.1.security-considerations.md
+++ b/drafts/use-cases/1.1.security-considerations.md
@@ -15,10 +15,14 @@ There are several possibilities here. One of them would be to use a *@graph* not
         {
             "@id": "/api",
             "@type": "hydra:EntryPoint",
-            "examplevocab:events": {
+            "collection": {
                 "@id": "/api/events",
                 "title": "List of events",
                 "@type": "hydra:Collection",
+                "manages": {
+                  "property": "rdf:type",
+                  "object": "schema:Event"
+                },
                 "operation": [
                     {
                        "@type": ["hydra:Operation", "schema:CreateAction"],
--- a/drafts/use-cases/1.entry-point.md
+++ b/drafts/use-cases/1.entry-point.md
@@ -34,10 +34,14 @@ HTTP 200 OK
     "@context": "/api/context.jsonld",
     "@id": "/api",
     "@type": "hydra:EntryPoint",
-    "examplevocab:events": {
+    "collection": {
         "@id": "/api/events",
         "title": "List of events",
         "@type": "hydra:Collection",
+        "manages": {
+          "property": "rdf:type",
+          "object": "schema:Event"
+        },
         "operation": [
             {
                 "@type": ["hydra:Operation", "schema:CreateAction"],
--- a/drafts/use-cases/3.obtaining-events.md
+++ b/drafts/use-cases/3.obtaining-events.md
@@ -35,6 +35,11 @@ HTTP 200 OK
 {
     "@context": "/api/context.jsonld",
     "@id": "/api/events",
+    "@type": "hydra:Collection",
+    "manages": {
+      "property": "rdf:type",
+      "object": "schema:Event"
+    },
     "totalItems": 1,
     "member": [
         {
--- a/drafts/use-cases/5.creating-new-event.md
+++ b/drafts/use-cases/5.creating-new-event.md
@@ -23,7 +23,10 @@ var client = new HydraClient();
 var operation = client.get("http://example.com")
     .getApiDocumentation()
     .getEntryPoint()
-    .getLink('http://example.com/vocab#events')
+    .getCollection({
+      property: "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
+      object: "http://schema.org/CreateEvent"
+      })
     .getOperationOfType('http://schema.org/CreateAction');
 client.invoke(operation, event);
 ```
--- a/drafts/use-cases/5.creating-new-event.md
+++ b/drafts/use-cases/5.creating-new-event.md
@@ -25,7 +25,7 @@ var operation = client.get("http://example.com")
     .getEntryPoint()
     .getCollection({
       property: "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
-      object: "http://schema.org/CreateEvent"
+      object: "http://schema.org/Event"
       })
     .getOperationOfType('http://schema.org/CreateAction');
 client.invoke(operation, event);
--- a/drafts/use-cases/7.searching-events.md
+++ b/drafts/use-cases/7.searching-events.md
@@ -48,6 +48,11 @@ HTTP 200 OK
 {
     "@context": "/api/context.jsonld",
     "@id": "/api/events",
+    "@type": "hydra:Collection",
+    "manages": {
+      "property": "rdf:type",
+      "object": "schema:Event"
+    },
     "totalItems": 1,
     "members": [
         {
@@ -86,6 +91,7 @@ HTTP 200 OK
 {
     "@context": "/api/context.jsonld",
     "@id": "/api/events?search=some text",
+    "@type": "hydra:Collection",
     "totalItems": 1,
     "members": [
         {
--- a/drafts/use-cases/8.filtering-events.md
+++ b/drafts/use-cases/8.filtering-events.md
@@ -51,6 +51,11 @@ HTTP 200 OK
 {
     "@context": "/api/context.jsonld",
     "@id": "/api/events",
+    "@type": "hydra:Collection",
+    "manages": {
+      "property": "rdf:type",
+      "object": "schema:Event"
+    },
     "totalItems": 1,
     "members": [
         {
@@ -89,6 +94,7 @@ HTTP 200 OK
 {
     "@context": "/api/context.jsonld",
     "@id": " /api/events?schema:eventName=.*test.*",
+    "@type": "hydra:Collection",
     "totalItems": 1,
     "members": [
         {


elf-pavlik:
Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion.

---

*[drafts/use-cases/5.creating-new-event.md, line 29 at r1](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrT4ZeUK43qWAnpOI76:-KrT4ZeUK43qWAnpOI77:bjvbxe0) ([raw file](https://github.com/hydracg/specifications/blob/7a85c04e5cbfa0958d3ffd64bb3fafbb423a1361/drafts/use-cases/5.creating-new-event.md#L29)):*
> ```Markdown
>       property: "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
>       object: "http://schema.org/CreateEvent"
>       })
> ```

client could have more specific method for such common use case, something like `getCollectionByMemberType('http://schema.org/Event')`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

tpluscode:
Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion.

---

*[drafts/use-cases/5.creating-new-event.md, line 29 at r1](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrT4ZeUK43qWAnpOI76:-KrURMBi0zq4P1QpeIJi:bfjq7x5) ([raw file](https://github.com/hydracg/specifications/blob/7a85c04e5cbfa0958d3ffd64bb3fafbb423a1361/drafts/use-cases/5.creating-new-event.md#L29)):*
<details><summary><i>Previously, elf-pavlik (elf Pavlik) wrote…</i></summary><blockquote>

client could have more specific method for such common use case, something like `getCollectionByMemberType('http://schema.org/Event')`
</blockquote></details>

I don't understand. You completely replaced the following of `#events` link with the `getCollection` call? In this case how does the client know which URL to retrieve?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

tpluscode:
Reviewed 2 of 7 files at r1.
Review status: 2 of 7 files reviewed at latest revision, 4 unresolved discussions.

---

*[drafts/use-cases/1.1.security-considerations.md, line 18 at r2](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrUSFWcxbytETQ1yIEc:-KrUSFWcxbytETQ1yIEd:b56qa6n) ([raw file](https://github.com/hydracg/specifications/blob/8c2db652e54f3e4ed52b855e05c73330d9d2d2ea/drafts/use-cases/1.1.security-considerations.md#L18)):*
> ```Markdown
>             "@id": "/api",
>             "@type": "hydra:EntryPoint",
>             "collection": {
> ```

Are you sure we should be changing this to `hydra:collection`? I guess this change answers my previous questions. But, I don't like how it loses the API-scecificness.

---

*[drafts/use-cases/7.searching-events.md, line 94 at r2](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrUTEC8cxSsvr6wpaDZ:-KrUTEC8cxSsvr6wpaD_:b-hg6vaa) ([raw file](https://github.com/hydracg/specifications/blob/8c2db652e54f3e4ed52b855e05c73330d9d2d2ea/drafts/use-cases/7.searching-events.md#L94)):*
> ```Markdown
>     "@context": "/api/context.jsonld",
>     "@id": "/api/events?search=some text",
>     "@type": "hydra:Collection",
> ```

Is this really a collection or a partial view?

---

*[drafts/use-cases/8.filtering-events.md, line 97 at r2](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrUTKKWRIulTHryZKEN:-KrUTKKWRIulTHryZKEO:b3hyiz6) ([raw file](https://github.com/hydracg/specifications/blob/8c2db652e54f3e4ed52b855e05c73330d9d2d2ea/drafts/use-cases/8.filtering-events.md#L97)):*
> ```Markdown
>     "@context": "/api/context.jsonld",
>     "@id": " /api/events?schema:eventName=.*test.*",
>     "@type": "hydra:Collection",
> ```

Ditto, Is this really a collection or a partial view?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

asbjornu:
Why are we using Reviewable for code reviews? I much prefer the code review feature of GitHub, tbh.

elf-pavlik:
@asbjornu plase see http://www.hydra-cg.com/minutes/2017-07-10/#resolution-1 (you will need to scroll up few lines)

elf-pavlik:
Review status: 2 of 7 files reviewed at latest revision, 4 unresolved discussions.

---

*[drafts/use-cases/1.1.security-considerations.md, line 18 at r2](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrUSFWcxbytETQ1yIEc:-KrVnkbjIm6WCU_qLzpF:bk6alry) ([raw file](https://github.com/hydracg/specifications/blob/8c2db652e54f3e4ed52b855e05c73330d9d2d2ea/drafts/use-cases/1.1.security-considerations.md#L18)):*
Could you please explain what you mean by:
> But, I don't like how it loses the API-scecificness.

To my understanding clients need a way to discover from the `hydra:EntryPoint` collection with members of type `schema:Event`.  But this general conversation already happens in https://github.com/HydraCG/Specifications/issues/126

---

*[drafts/use-cases/5.creating-new-event.md, line 29 at r1](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrT4ZeUK43qWAnpOI76:-KrVmtzhntSVrK3lTxKk:bkae26r) ([raw file](https://github.com/hydracg/specifications/blob/7a85c04e5cbfa0958d3ffd64bb3fafbb423a1361/drafts/use-cases/5.creating-new-event.md#L29)):*
<details><summary><i>Previously, tpluscode (Tomasz Pluskiewicz) wrote…</i></summary><blockquote>

I don't understand. You completely replaced the following of `#events` link with the `getCollection` call? In this case how does the client know which URL to retrieve?
</blockquote></details>

`@id` of the matching `hydra:Collection` I haven't thought of a case where multiple ones match but AFAIK `hydra:Link` also doesn't have restriction to reference only one resource

---

*[drafts/use-cases/7.searching-events.md, line 94 at r2](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrUTEC8cxSsvr6wpaDZ:-KrVlhLXqKM0BVudpPqz:b-lxdirj) ([raw file](https://github.com/hydracg/specifications/blob/8c2db652e54f3e4ed52b855e05c73330d9d2d2ea/drafts/use-cases/7.searching-events.md#L94)):*
<details><summary><i>Previously, tpluscode (Tomasz Pluskiewicz) wrote…</i></summary><blockquote>

Is this really a collection or a partial view?
</blockquote></details>

Good question, I hope Markus and Karol can chime in on that. I could always remove this change from this PR since it doesn't really belong in it.

---

*[drafts/use-cases/8.filtering-events.md, line 97 at r2](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrUTKKWRIulTHryZKEN:-KrVx6rbzWUu1Q1M2vt0:b-3k7s37) ([raw file](https://github.com/hydracg/specifications/blob/8c2db652e54f3e4ed52b855e05c73330d9d2d2ea/drafts/use-cases/8.filtering-events.md#L97)):*
<details><summary><i>Previously, tpluscode (Tomasz Pluskiewicz) wrote…</i></summary><blockquote>

Ditto, Is this really a collection or a partial view?
</blockquote></details>

I could also undo this change and leave it for another PR

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

tpluscode:
Reviewed 4 of 7 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[drafts/use-cases/1.1.security-considerations.md, line 18 at r2](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrUSFWcxbytETQ1yIEc:-KrWjnr3sOtOBzyBPQqw:b-6whcbg) ([raw file](https://github.com/hydracg/specifications/blob/8c2db652e54f3e4ed52b855e05c73330d9d2d2ea/drafts/use-cases/1.1.security-considerations.md#L18)):*
<details><summary><i>Previously, elf-pavlik (elf Pavlik) wrote…</i></summary><blockquote>

Could you please explain what you mean by:
> But, I don't like how it loses the API-scecificness.

To my understanding clients need a way to discover from the `hydra:EntryPoint` collection with members of type `schema:Event`.  But this general conversation already happens in https://github.com/HydraCG/Specifications/issues/126
</blockquote></details>

What I mean is that replacing the API-specific term loses the link type between one resource (the Entrypoint) and the other (List of events). I think that the link relation trumps the manages block, so to speak.

As for the client though, a way to discover collection of a given type may be useful. How would you like it to be based on `"@type": "hydra:Collection"` and the `manages`-block and not the predicate type? This way you can keep a more informative link predicate

---

*[drafts/use-cases/5.creating-new-event.md, line 29 at r1](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrT4ZeUK43qWAnpOI76:-KrWl6u74vFFICdRFAsm:b-1ow12t) ([raw file](https://github.com/hydracg/specifications/blob/7a85c04e5cbfa0958d3ffd64bb3fafbb423a1361/drafts/use-cases/5.creating-new-event.md#L29)):*
<details><summary><i>Previously, elf-pavlik (elf Pavlik) wrote…</i></summary><blockquote>

`@id` of the matching `hydra:Collection` I haven't thought of a case where multiple ones match but AFAIK `hydra:Link` also doesn't have restriction to reference only one resource
</blockquote></details>

Not really what I had in mind but you are right. I meant more of what I wrote in the other comment about losing the link type. 

In the end though I still don't entirely understand why you'd want a `getCollection` method instead of `getLink`. The former assumes what's on the other side. Not really hypermedia-y.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

tpluscode:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[drafts/use-cases/1.1.security-considerations.md, line 18 at r2](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrUSFWcxbytETQ1yIEc:-KrWvvCnjeEYb2_wB-YC:blormog) ([raw file](https://github.com/hydracg/specifications/blob/8c2db652e54f3e4ed52b855e05c73330d9d2d2ea/drafts/use-cases/1.1.security-considerations.md#L18)):*
<details><summary><i>Previously, lanthaler (Markus Lanthaler) wrote…</i></summary><blockquote>

If such a relation exists, then yes, it should be used. This is for cases where no such relations exists yet and would only be created for the sake of being able to connect some entity with a collection.
</blockquote></details>

Hm, wouldn't it be a good rule of thumb to mint an API-specific property for any such case? I think that is precisely what you did in events example...

---

*[drafts/use-cases/5.creating-new-event.md, line 29 at r1](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrT4ZeUK43qWAnpOI76:-KrWwfKH-Fadrl8xtVZ4:b-3sunog) ([raw file](https://github.com/hydracg/specifications/blob/7a85c04e5cbfa0958d3ffd64bb3fafbb423a1361/drafts/use-cases/5.creating-new-event.md#L29)):*
<details><summary><i>Previously, lanthaler (Markus Lanthaler) wrote…</i></summary><blockquote>

Think of it as a specialized link relation that Hydra clients understand natively. This is similar to how browsers use the `icon` or `stylesheet` link relation. You could also have multiple `manages` blocks that describe the relationship between the referencing entity and the collection members in more detail. See the [page on the wiki](https://www.w3.org/community/hydra/wiki/Avoid_that_collections_break_relationships) where we worked on this (here's the [summary](https://www.w3.org/community/hydra/wiki/Collection_Design)).
</blockquote></details>

The problem with you analogy is that icon clearly defines the contents of the link's target. I'm not sure that is the case here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

lanthaler:
Reviewed 1 of 7 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[drafts/use-cases/1.1.security-considerations.md, line 18 at r2](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrUSFWcxbytETQ1yIEc:-KrWxriD2rz8IzgZP7iz:b-ci6iv9) ([raw file](https://github.com/hydracg/specifications/blob/8c2db652e54f3e4ed52b855e05c73330d9d2d2ea/drafts/use-cases/1.1.security-considerations.md#L18)):*
<details><summary><i>Previously, tpluscode (Tomasz Pluskiewicz) wrote…</i></summary><blockquote>

Hm, wouldn't it be a good rule of thumb to mint an API-specific property for any such case? I think that is precisely what you did in events example...
</blockquote></details>

Yeah, it's exactly what I did in the events example.. but only because there was nothing else available. The goal is to create clients that "understand" these APIs to some degree so that they can figure out how to use them. If everyone mints property like `myvocab:events` only clients that have been written specifically for that API will understand them. If we have a generic mechanisms as the one Pavlik proposes here on the other hand, we can implement client libraries that can *find the collection that contains all events*.

A custom link relation is only useful if it has some well-defined semantics that you can't convey otherwise.

---

*[drafts/use-cases/5.creating-new-event.md, line 29 at r1](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrT4ZeUK43qWAnpOI76:-KrWyhHk7NaD-nzLF9Uc:b2a66mm) ([raw file](https://github.com/hydracg/specifications/blob/7a85c04e5cbfa0958d3ffd64bb3fafbb423a1361/drafts/use-cases/5.creating-new-event.md#L29)):*
<details><summary><i>Previously, tpluscode (Tomasz Pluskiewicz) wrote…</i></summary><blockquote>

The problem with you analogy is that icon clearly defines the contents of the link's target. I'm not sure that is the case here.
</blockquote></details>

[Nope](https://www.w3.org/TR/html5/links.html#rel-icon), unless I misunderstand you. It just describes in what relationship the target (the icon) stands to the context IRI (the document). The optional target attributes on a link are just hints that help a client to choose the supposedly best matching link. It must not trust them though but interpret what it actually gets when it dereferences the URL.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

elf-pavlik:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[drafts/use-cases/0.intro.md, line 89 at r2](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrWt32Z1UxZkNXLNAXm:-KrXdrB95u6yx22c-D0f:byvigx2) ([raw file](https://github.com/hydracg/specifications/blob/8c2db652e54f3e4ed52b855e05c73330d9d2d2ea/drafts/use-cases/0.intro.md#L89)):*
<details><summary><i>Previously, lanthaler (Markus Lanthaler) wrote…</i></summary><blockquote>

I don't remember whether we introduced the rule afterwards but I think we agreed to not have terms that differ only in their capitalization. Maybe we should revisit this at some point.
</blockquote></details>

captured in https://github.com/HydraCG/Specifications/issues/133

---

*[drafts/use-cases/1.1.security-considerations.md, line 18 at r2](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrUSFWcxbytETQ1yIEc:-KrXjKLiBHWVcSKkLQQ5:b7f3y0o) ([raw file](https://github.com/hydracg/specifications/blob/8c2db652e54f3e4ed52b855e05c73330d9d2d2ea/drafts/use-cases/1.1.security-considerations.md#L18)):*
<details><summary><i>Previously, lanthaler (Markus Lanthaler) wrote…</i></summary><blockquote>

Yeah, it's exactly what I did in the events example.. but only because there was nothing else available. The goal is to create clients that "understand" these APIs to some degree so that they can figure out how to use them. If everyone mints property like `myvocab:events` only clients that have been written specifically for that API will understand them. If we have a generic mechanisms as the one Pavlik proposes here on the other hand, we can implement client libraries that can *find the collection that contains all events*.

A custom link relation is only useful if it has some well-defined semantics that you can't convey otherwise.
</blockquote></details>

Since this PR proposes just one of possible solutions for the problem, I would like to suggest once again discussing and further clarifying the problem we try to address itself in https://github.com/HydraCG/Specifications/issues/126#issuecomment-322320409 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

tpluscode:
Reviewed 1 of 7 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[drafts/use-cases/1.1.security-considerations.md, line 18 at r2](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrUSFWcxbytETQ1yIEc:-Kr_2TJ5W_aCIhX9duPe:bv2279w) ([raw file](https://github.com/hydracg/specifications/blob/8c2db652e54f3e4ed52b855e05c73330d9d2d2ea/drafts/use-cases/1.1.security-considerations.md#L18)):*
<details><summary><i>Previously, elf-pavlik (elf Pavlik) wrote…</i></summary><blockquote>

Since this PR proposes just one of possible solutions for the problem, I would like to suggest once again discussing and further clarifying the problem we try to address itself in https://github.com/HydraCG/Specifications/issues/126#issuecomment-322320409 
</blockquote></details>

@mlanthaler only clients written for a specific API will understand the contents of the given collection anyway. the will understand `hydra:Collection` and partial views. other than that I don't see how having the collection is really useful as fallback. won't this lead to

``` json
{
  "@id": "/some-person"
  "hydra:collection": [
    {
      "@id": "/some-person/friends"
    },
    {
      "@id": "/some-person/enemies"
    },
    {
      "@id": "/some-person/interests"
    }
  ]
}
```

how useful is it now? Also, this PR assumes that the content of a collection is inline. If you only got the target URL, the `hydra:collection` conveys completely no information.

---

*[drafts/use-cases/5.creating-new-event.md, line 29 at r1](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrT4ZeUK43qWAnpOI76:-Kr_3pYuM-3ERYqQHOAI:bemgsjd) ([raw file](https://github.com/hydracg/specifications/blob/7a85c04e5cbfa0958d3ffd64bb3fafbb423a1361/drafts/use-cases/5.creating-new-event.md#L29)):*
<details><summary><i>Previously, lanthaler (Markus Lanthaler) wrote…</i></summary><blockquote>

[Nope](https://www.w3.org/TR/html5/links.html#rel-icon), unless I misunderstand you. It just describes in what relationship the target (the icon) stands to the context IRI (the document). The optional target attributes on a link are just hints that help a client to choose the supposedly best matching link. It must not trust them though but interpret what it actually gets when it dereferences the URL.
</blockquote></details>

Well, my point precisely (see the comment above). If the representation contained multiple `hydra:collection`links to multiple unrelated resources, the client would not be able to determine which one to dereference. Unless they are meant to be direct alternatives. Unless the `manages`-block will be mandatory...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

elf-pavlik:
*[drafts/use-cases/1.1.security-considerations.md, line 18 at r2](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrUSFWcxbytETQ1yIEc:-Kr_oj1Yn4ek-c68NOX_:brchfta) ([raw file](https://github.com/hydracg/specifications/blob/8c2db652e54f3e4ed52b855e05c73330d9d2d2ea/drafts/use-cases/1.1.security-considerations.md#L18)):*
> only clients written for a specific API will understand the contents of the given collection anyway.

I think we need to disambiguate between *vocabs* (schemas), *APIs* and *datasets*. Preferably around use cases included in our repo. In case where particular client wants to interact with **multiple** servers, each of those servers exposing a dataset via API but all those datasets use a common vocab (eg. schema.org). Preferably client only needs to implement Hydra spec and understand schema.org vocab to interact with all those servers exposing all those datasets.

If any of those servers defines some custom `hydra:Link` it can't expect such generic client to understand it. In our case relating collection of events from entry point with something as simple as `rdfs:seeAlso` and using `hydra:manages` block to hint type of all the members seems sufficient for client looking for 'collection of events in this dataset'

> Also, this PR assumes that the content of a collection is inline. If you only got the target URL, the hydra:collection conveys completely no information.

I think while discussing collection design we have considered using `rdfs:seeAlso` but then decided to add `hydra:collection` to the vocab. I would see this property (the information it conveys) as something close to `rdfs:seeAlso` and 'void:subset'.

> won't this lead to { ... } how useful is it now?

For example [Solid mandates](https://github.com/solid/solid-spec/blob/master/solid-webid-profiles.md#extended-profile) clients MUST load and parse all resources linked from personal WebID profile to get combined 'extended profile'. Clients which don't find enough information about collection inlined in entry point, can simply fetch it. We should still encourage publishers to inline information relevant for discovery in entry point.
BTW for the *friends*, *enemies*, *interests* manages block still seems more practical than trying to make up some `hydra:Link`instances
* http://purl.org/vocab/relationship/friendOf
* http://purl.org/vocab/relationship/enemyOf
* http://smiy.sourceforge.net/cco/spec/cognitivecharacteristics.html#interest

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

tpluscode:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[drafts/use-cases/1.1.security-considerations.md, line 18 at r2](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrUSFWcxbytETQ1yIEc:-KraHnt_nrVC66Hc7fyf:b-gpmnig) ([raw file](https://github.com/hydracg/specifications/blob/8c2db652e54f3e4ed52b855e05c73330d9d2d2ea/drafts/use-cases/1.1.security-considerations.md#L18)):*
<details><summary><i>Previously, elf-pavlik (elf Pavlik) wrote…</i></summary><blockquote>

> only clients written for a specific API will understand the contents of the given collection anyway.

I think we need to disambiguate between *vocabs* (schemas), *APIs* and *datasets*. Preferably around use cases included in our repo. In case where particular client wants to interact with **multiple** servers, each of those servers exposing a dataset via API but all those datasets use a common vocab (eg. schema.org). Preferably client only needs to implement Hydra spec and understand schema.org vocab to interact with all those servers exposing all those datasets.

If any of those servers defines some custom `hydra:Link` it can't expect such generic client to understand it. In our case relating collection of events from entry point with something as simple as `rdfs:seeAlso` and using `hydra:manages` block to hint type of all the members seems sufficient for client looking for 'collection of events in this dataset'

> Also, this PR assumes that the content of a collection is inline. If you only got the target URL, the hydra:collection conveys completely no information.

I think while discussing collection design we have considered using `rdfs:seeAlso` but then decided to add `hydra:collection` to the vocab. I would see this property (the information it conveys) as something close to `rdfs:seeAlso` and 'void:subset'.

> won't this lead to { ... } how useful is it now?

For example [Solid mandates](https://github.com/solid/solid-spec/blob/master/solid-webid-profiles.md#extended-profile) clients MUST load and parse all resources linked from personal WebID profile to get combined 'extended profile'. Clients which don't find enough information about collection inlined in entry point, can simply fetch it. We should still encourage publishers to inline information relevant for discovery in entry point.
BTW for the *friends*, *enemies*, *interests* manages block still seems more practical than trying to make up some `hydra:Link`instances
* http://purl.org/vocab/relationship/friendOf
* http://purl.org/vocab/relationship/enemyOf
* http://smiy.sourceforge.net/cco/spec/cognitivecharacteristics.html#interest

</blockquote></details>

The element type of `friends` and `enemies` will both be a Person in the manages block? Again, the client cannot distinguish between two collection having clearly different meaning...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

elf-pavlik:
*[drafts/use-cases/1.1.security-considerations.md, line 18 at r2](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrUSFWcxbytETQ1yIEc:-KrafHHAVaZSxiDYZx8V:ba321m7) ([raw file](https://github.com/hydracg/specifications/blob/8c2db652e54f3e4ed52b855e05c73330d9d2d2ea/drafts/use-cases/1.1.security-considerations.md#L18)):*
<details><summary><i>Previously, tpluscode (Tomasz Pluskiewicz) wrote…</i></summary><blockquote>

The element type of `friends` and `enemies` will both be a Person in the manages block? Again, the client cannot distinguish between two collection having clearly different meaning...
</blockquote></details>

In case above manages blocks wouldn't rely on 'type' `property: rdf:type` but instead on properties like:

  *  http://purl.org/vocab/relationship/friendOf
  * http://purl.org/vocab/relationship/enemyOf
  * http://smiy.sourceforge.net/cco/spec/cognitivecharacteristics.html#interest

Which  capture the relation between `/some-person` and each member in the collection. Not the 'type' of the member -  relation between some instance of `rdfs:Class` and each member in the collection.

Have you participated in defining [Collection Design](https://www.w3.org/community/hydra/wiki/Collection_Design)? Possibly it needs better documentation and clarified motivation beyond linking to gh issues. We didn't think about use case of `rdf:type` property when we discussed his design but it seems perfectly fit to handle it as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

tpluscode:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[drafts/use-cases/1.1.security-considerations.md, line 18 at r2](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrUSFWcxbytETQ1yIEc:-KrbKf_ipQMlw5cMwIJn:byqtme2) ([raw file](https://github.com/hydracg/specifications/blob/8c2db652e54f3e4ed52b855e05c73330d9d2d2ea/drafts/use-cases/1.1.security-considerations.md#L18)):*
<details><summary><i>Previously, elf-pavlik (elf Pavlik) wrote…</i></summary><blockquote>

In case above manages blocks wouldn't rely on 'type' `property: rdf:type` but instead on properties like:

  *  http://purl.org/vocab/relationship/friendOf
  * http://purl.org/vocab/relationship/enemyOf
  * http://smiy.sourceforge.net/cco/spec/cognitivecharacteristics.html#interest

Which  capture the relation between `/some-person` and each member in the collection. Not the 'type' of the member -  relation between some instance of `rdfs:Class` and each member in the collection.

Have you participated in defining [Collection Design](https://www.w3.org/community/hydra/wiki/Collection_Design)? Possibly it needs better documentation and clarified motivation beyond linking to gh issues. We didn't think about use case of `rdf:type` property when we discussed his design but it seems perfectly fit to handle it as well.
</blockquote></details>

Ah, I think I recall. This is for when one resource doesn't have an actual relation in dataset but only in hypermedia. So in RDF you don't have a `api:entrypoint vocab:friends api:collection` triple but want to be able to include a link between the two in Hydra representation. Correct?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

lanthaler:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[drafts/use-cases/0.intro.md, line 89 at r2](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrWt32Z1UxZkNXLNAXm:-KrbPsgUmiZQKve7_j6J:b-zcjvte) ([raw file](https://github.com/hydracg/specifications/blob/8c2db652e54f3e4ed52b855e05c73330d9d2d2ea/drafts/use-cases/0.intro.md#L89)):*
<details><summary><i>Previously, elf-pavlik (elf Pavlik) wrote…</i></summary><blockquote>

captured in https://github.com/HydraCG/Specifications/issues/133
</blockquote></details>

Thanks

---

*[drafts/use-cases/5.creating-new-event.md, line 29 at r1](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrT4ZeUK43qWAnpOI76:-KrbQuq-JrSWABguh6EA:bapmyay) ([raw file](https://github.com/hydracg/specifications/blob/7a85c04e5cbfa0958d3ffd64bb3fafbb423a1361/drafts/use-cases/5.creating-new-event.md#L29)):*
<details><summary><i>Previously, tpluscode (Tomasz Pluskiewicz) wrote…</i></summary><blockquote>

Well, my point precisely (see the comment above). If the representation contained multiple `hydra:collection`links to multiple unrelated resources, the client would not be able to determine which one to dereference. Unless they are meant to be direct alternatives. Unless the `manages`-block will be mandatory...
</blockquote></details>

Correct, you do need the manages block to differentiate them. The advantage is that it allows you to describe things without having to invent new concepts. Assuming `X` points to `Y` via `hydra:collection`, manages can be used to express that 
 * all entries in `Y` are of a specific type:
    `"manages": { "property": "rdf:type", "object": "schema:Event" }`
 * all entries in `Y`stand in a specific relationship with `X`: 
    `"manages": { "subject": "X", "property": "schema:knows" }`
 * all entries in `Y` have a a specific property set to a specific value:
    `"manages": { "property": "schema:gender", "object": "schema:Female" }`

We can even allow multiple of them at the same time, so we would be able to express that there's a collection which manages entities of type `schema:Person`, that `X` `schema:knowns` and have `schema:gender``schema:Female`; in other words a collection containing `X`'s female acquaintances. That's pretty powerful.

Is that concern here actually that it looks more complex/is more verbose or that it has weaker semantics? I can agree with the former but don't think the latter is true.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

elf-pavlik:
*[drafts/use-cases/1.1.security-considerations.md, line 18 at r2](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrUSFWcxbytETQ1yIEc:-KrezLD2KNi1RKT_pm4b:b2v3y85) ([raw file](https://github.com/hydracg/specifications/blob/8c2db652e54f3e4ed52b855e05c73330d9d2d2ea/drafts/use-cases/1.1.security-considerations.md#L18)):*
<details><summary><i>Previously, tpluscode (Tomasz Pluskiewicz) wrote…</i></summary><blockquote>

Ah, I think I recall. This is for when one resource doesn't have an actual relation in dataset but only in hypermedia. So in RDF you don't have a `api:entrypoint vocab:friends api:collection` triple but want to be able to include a link between the two in Hydra representation. Correct?
</blockquote></details>

Not sure if I understood. In case `/some-person` of `socrel:friendOf` or 'cco:interest` relations exist in the dataset, not between the `/some-person` and`hydra:collection` but between `/some-person` and each member. In case of `rdf:type` no relation meaningful for the dataset seems to exist between each `schema:Event` and the `hydra:EntryPoint` (except something like `void:inDataset`). If client would use SPARQL or Triple Pattern Fragments interface, the `hydra:collection` seems not needed and IMO often may come just as artifact of trying to provide some tree-ish layout for REST inteface to access the graph-y dataset. I think to really clarify that, we have to keep working with more and more use cases and for each one take a look how accessing the same dataset would differ if using SPARQL interface or HyDRA interface - which data we add just for the sake of creating HyDRA interface layout.
For the sake of this PR I think that Markus gave really good answer in another thread, the one with examples like *all entries in Y have a a specific property set to a specific value:*. Main improvement in this PR comes with switching to generalized approach relying on already existing terms in common vocabs and nod needing to improvise terms like `mysnowflakevocab:events` which go against benefits of *shared understanding* provided by commonly used vocabs and enabling interoperability.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

tpluscode:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[drafts/use-cases/5.creating-new-event.md, line 29 at r1](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrT4ZeUK43qWAnpOI76:-KrgNBslzvwC-FuOm8X1:b-yi0we3) ([raw file](https://github.com/hydracg/specifications/blob/7a85c04e5cbfa0958d3ffd64bb3fafbb423a1361/drafts/use-cases/5.creating-new-event.md#L29)):*
<details><summary><i>Previously, lanthaler (Markus Lanthaler) wrote…</i></summary><blockquote>

Correct, you do need the manages block to differentiate them. The advantage is that it allows you to describe things without having to invent new concepts. Assuming `X` points to `Y` via `hydra:collection`, manages can be used to express that 
 * all entries in `Y` are of a specific type:
    `"manages": { "property": "rdf:type", "object": "schema:Event" }`
 * all entries in `Y`stand in a specific relationship with `X`: 
    `"manages": { "subject": "X", "property": "schema:knows" }`
 * all entries in `Y` have a a specific property set to a specific value:
    `"manages": { "property": "schema:gender", "object": "schema:Female" }`

We can even allow multiple of them at the same time, so we would be able to express that there's a collection which manages entities of type `schema:Person`, that `X` `schema:knowns` and have `schema:gender``schema:Female`; in other words a collection containing `X`'s female acquaintances. That's pretty powerful.

Is that concern here actually that it looks more complex/is more verbose or that it has weaker semantics? I can agree with the former but don't think the latter is true.
</blockquote></details>

My concern is that this looks verbose and complicated. Especially that this structure addresses a RDF-specific problem. But, I don't have any more comments. Thank you for your patience :D

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

asbjornu:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[drafts/use-cases/5.creating-new-event.md, line 29 at r1](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrT4ZeUK43qWAnpOI76:-Krgo--7nhoYDX4RIt7U:b-j64585) ([raw file](https://github.com/hydracg/specifications/blob/7a85c04e5cbfa0958d3ffd64bb3fafbb423a1361/drafts/use-cases/5.creating-new-event.md#L29)):*
<details><summary><i>Previously, tpluscode (Tomasz Pluskiewicz) wrote…</i></summary><blockquote>

My concern is that this looks verbose and complicated. Especially that this structure addresses a RDF-specific problem. But, I don't have any more comments. Thank you for your patience :D
</blockquote></details>

I agree with Tomasz that this looks complicated. With my lack of RDF knowledge, I hardly understand what's going on. :confused: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

RubenVerborgh:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[drafts/use-cases/5.creating-new-event.md, line 29 at r1](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrT4ZeUK43qWAnpOI76:-KrgxZAN0v-H30HiV02W:bj3w62y) ([raw file](https://github.com/hydracg/specifications/blob/7a85c04e5cbfa0958d3ffd64bb3fafbb423a1361/drafts/use-cases/5.creating-new-event.md#L29)):*
<details><summary><i>Previously, asbjornu (Asbjørn Ulsberg) wrote…</i></summary><blockquote>

I agree with Tomasz that this looks complicated. With my lack of RDF knowledge, I hardly understand what's going on. :confused: 
</blockquote></details>

Can't `getCollection` take JSON-LD (with an implicit or explicit context)? Then we can write this much more clearly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

elf-pavlik:
*[drafts/use-cases/5.creating-new-event.md, line 29 at r1](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrT4ZeUK43qWAnpOI76:-KrhkS2_EClBffiyUoqf:b20hu7i) ([raw file](https://github.com/hydracg/specifications/blob/7a85c04e5cbfa0958d3ffd64bb3fafbb423a1361/drafts/use-cases/5.creating-new-event.md#L29)):*
<details><summary><i>Previously, RubenVerborgh (Ruben Verborgh) wrote…</i></summary><blockquote>

Can't `getCollection` take JSON-LD (with an implicit or explicit context)? Then we can write this much more clearly.
</blockquote></details>

I think the client should provide interface to control JSON-LD context. We just need to make it clear that developer shouldn't make any assumptions about JSON-LD context in received payloads and aways make sure to provide desired context to the client.
I would propose that once we agree on using `manages` block in the use case we should address client interface details in a future PR to https://github.com/HydraCG/Heracles.ts which implementation follows work on use cases...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

lanthaler:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[drafts/use-cases/5.creating-new-event.md, line 29 at r1](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrT4ZeUK43qWAnpOI76:-KriuXCk_uMimMoTGeQs:bbe5jpo) ([raw file](https://github.com/hydracg/specifications/blob/7a85c04e5cbfa0958d3ffd64bb3fafbb423a1361/drafts/use-cases/5.creating-new-event.md#L29)):*
<details><summary><i>Previously, elf-pavlik (elf Pavlik) wrote…</i></summary><blockquote>

I think the client should provide interface to control JSON-LD context. We just need to make it clear that developer shouldn't make any assumptions about JSON-LD context in received payloads and aways make sure to provide desired context to the client.
I would propose that once we agree on using `manages` block in the use case we should address client interface details in a future PR to https://github.com/HydraCG/Heracles.ts which implementation follows work on use cases...
</blockquote></details>

Yeah, I share that concerns as we "leak" quite a lot of RDF concepts here. Perhaps renaming `manages` to `constraints` or something similar and trying to come up with something replacing `subject` and `object` would solve that (we could eventually look at [SHACL](https://www.w3.org/TR/shacl/) which recently became a W3C recommendation).

In the spirit of moving fast(er) I'd, howerver, like to propose to go ahead and merge this PR. If someone has a concrete proposal to improve this, it can be discussed in a separate PR. If no one objects, I'll merge this tomorrow.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

alien-mcl:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[drafts/use-cases/5.creating-new-event.md, line 29 at r1](https://reviewable.io:443/reviews/hydracg/specifications/132#-KrT4ZeUK43qWAnpOI76:-Ks4yEH8dvXZCcUP-BW0:b-6klco5) ([raw file](https://github.com/hydracg/specifications/blob/7a85c04e5cbfa0958d3ffd64bb3fafbb423a1361/drafts/use-cases/5.creating-new-event.md#L29)):*
<details><summary><i>Previously, lanthaler (Markus Lanthaler) wrote…</i></summary><blockquote>

Yeah, I share that concerns as we "leak" quite a lot of RDF concepts here. Perhaps renaming `manages` to `constraints` or something similar and trying to come up with something replacing `subject` and `object` would solve that (we could eventually look at [SHACL](https://www.w3.org/TR/shacl/) which recently became a W3C recommendation).

In the spirit of moving fast(er) I'd, howerver, like to propose to go ahead and merge this PR. If someone has a concrete proposal to improve this, it can be discussed in a separate PR. If no one objects, I'll merge this tomorrow.
</blockquote></details>

Tha major issue I've got with `getCollection` is that the client need's to know ad-hoc that the result is a collection.

As for the JSON-LD filtering - I agree that we shouldn't be JSON-LD bound.

And yes - it seems like `filtering`. Do we have an agreement here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hydracg/specifications/132)*

Issue #42, opened 2014-04-17
Rethink paging


RubenVerborgh:
As [discussed on the mailing list](http://lists.w3.org/Archives/Public/public-hydra/2014Apr/0034.html), the current implementation of paging might be confusing. This became apparent when I was implementing this for [the Linked Data Fragments server](http://data.linkeddatafragments.org/dbpedia) (try the Turtle or JSON-LD representaiton).

Basically, the whole misconception for me comes with the term `PagedCollection`. If I told you "x is a paged collection", what would you expect? That x is a single page or that x is a collection that is accessible as pages?
Hydra says that a `PagedCollection` is  a page, as is apparent from the current spec:

```javascript
// EXAMPLE 14: A Hydra PagedCollection splits a collection into multiple pages
{
  "@context": "http://www.w3.org/ns/hydra/context.jsonld",
  "@id": "http://api.example.com/an-issue/comments?page=3",
  "@type": "PagedCollection",
  "totalItems": "4980",
  "itemsPerPage": "10",
  "firstPage": "/an-issue/comments?page=1",
  "nextPage": "/an-issue/comments?page=4",
  "previousPage": "/an-issue/comments?page=2",
  "lastPage": "/an-issue/comments?page=498",
  "member": [

  ]
}
```

This seems like a misnomer. What can we do to fix? Suggestions in comments.

---
BTW @lanthaler I would be totally willing to write this into the spec.

lanthaler:
Proposals are being tracked at https://www.w3.org/community/hydra/wiki/Pagination

lanthaler:
PROPOSAL: Look at pages of a collection as specific *views* on a single underlying collection instead of thinking of the collection as the sum of its pages. More concretely,
 - replace PagedCollection with PartialCollectionView,
 - replace firstPage/nextPage/previousPage/lastPage with first/next/previous/last,
 - associate totalItems with Collection instead of PagedCollection, and
 - remove itemsPerPage for the time being (we will likely re-introduce it with a different name).

lanthaler:
Call for consensus: https://lists.w3.org/Archives/Public/public-hydra/2015Oct/0163.html

lanthaler:
[...]

PR #165, opened 2018-05-13
Update UC of partial collection

### LINKED ISSUE #163: Inconsistent examples regarding pagination
The recently merged PR #159 added some new use case documents. The one for pagination embedded the `first`, `next`, etc links in the root of the collection document, while most other examples I've seen for pagination put them nested under a `view` key in a PartialCollectionView object. Which is the correct way? Should the examples at least be consistent?

https://github.com/HydraCG/Specifications/pull/159/files/1f4e6027d8885cdede04d300c5f5625aaf2c9fc3#diff-cc46591d2cf05932a3d79d5806d206dfR57

tpluscode:
Fixes #163

I've updated the example of a partial collection to be represented using the `PartialCollectionView` as agreed in #42.

### DIFF
--- a/drafts/use-cases/3.2.pagination.md
+++ b/drafts/use-cases/3.2.pagination.md
@@ -18,13 +18,13 @@ So I can start reviewing them with option of loading more on demand.
 
 ```javascript
 const client = new HydraClient();
-let data = client.get("/api/events");
-for (const member of data.members) {
+let collection = client.get("/api/events");
+for (const member of collection.members) {
     // do something with the _member_, i.e. display it
 }
 // load some more on demand
-data = client.get(data.view.next);
-for (const member of data.members) {
+collection = client.get(data.view.next);
+for (const member of collection.members) {
     // do something with more _member_, i.e. display it
 }
 ```
--- a/drafts/use-cases/3.2.pagination.md
+++ b/drafts/use-cases/3.2.pagination.md
@@ -23,7 +23,7 @@ for (const member of data.members) {
     // do something with the _member_, i.e. display it
 }
 // load some more on demand
-data = client.get(data.next);
+data = client.get(data.view.next);
 for (const member of data.members) {
     // do something with more _member_, i.e. display it
 }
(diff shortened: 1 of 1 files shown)


drafts/use-cases/3.2.pagination.md
```diff
@@ -18,13 +18,13 @@ So I can start reviewing them with option of loading more on demand.
...
-let data = client.get("/api/events");
-for (const member of data.members) {
+let collection = client.get("/api/events");
+for (const member of collection.members) {
     // do something with the _member_, i.e. display it
 }
 // load some more on demand
-data = client.get(data.next);
-for (const member of data.members) {
+collection = client.get(data.view.next);
```
tpluscode: @alien-mcl Maybe this snippet could already show the new API proposed in HydraCG/Heracles.ts#34

drafts/use-cases/3.2.pagination.md
```diff
@@ -18,13 +18,13 @@ So I can start reviewing them with option of loading more on demand.
...
-let data = client.get("/api/events");
-for (const member of data.members) {
+let collection = client.get("/api/events");
+for (const member of collection.members) {
     // do something with the _member_, i.e. display it
 }
 // load some more on demand
-data = client.get(data.next);
-for (const member of data.members) {
+collection = client.get(data.view[0].next);
```
tpluscode: @alien-mcl Maybe this snippet could already use the new API as proposed in HydraCG/Heracles.ts#34

tpluscode:
@lanthaler, do we still keep the second `items` property to give the number in a single view?

```json
{
  "totalItems": 100, // total
  "view": [{
    "items": 10 // this page/view
  }]
}
```

drafts/use-cases/3.2.pagination.md
alien-mcl: I think it's a good idea. I'll try to update that PR with API "release candidate" :)

lanthaler:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[drafts/use-cases/3.2.pagination.md, line 26 at r1](https://reviewable.io/reviews/hydracg/specifications/165#-LDWTlCz6Zn97qGJyrVM-r1-26:-LDWTyoJ3hTzsITVoqnY:blzautf) ([raw file](https://github.com/hydracg/specifications/blob/13c3c6fb962eea86fe7bf3b502ed4281a30d7c3b/drafts/use-cases/3.2.pagination.md#L26)):*
<details><summary><i>Previously, alien-mcl (Karol Szczepański) wrote…</i></summary><blockquote>

I think it's a good idea. I'll try to update that PR with API "release candidate" :)
</blockquote></details>

Yeah, just calling `next` here would be problematic as that would return the PartialCollectionView instead. Also, you renamed `data` to `collection` so you need to change it in this line as well

---


*Comments from [Reviewable](https://reviewable.io/reviews/hydracg/specifications/165)*

lanthaler:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io/reviews/hydracg/specifications/165)*

drafts/use-cases/3.2.pagination.md
vddesai1871: If I am correct we have also replaced `get` method with `getResource`. If no one is working on this then I would like to create a PR adapting use cases to new client API.

PR #174, opened 2019-01-08
Specification change to cover the issue

alien-mcl:
Added a proper paragraphs recommending dereferencing untyped links from the API's domain. No other vocabulary constructs were presented.

### DIFF
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -314,6 +314,11 @@
       <code>comments</code> property in the following example is intended
       to be dereferenced.</p>
 
+    <p class="note">It is recommended to dereference resources that are
+      within API's domain. This may prevent possible issues with cross-site
+      scripting or obtaining meaningless resources. Still, there is no formal
+      prohibition of dereferencing <i>untyped</i> links, i.e. <i>rdf:seeAlso</i>.</p>
+
     <pre class="example nohighlight" data-transform="updateExample"
          title="Using a property defined to be a hyperlink">
       <!--


spec/latest/core/index.html
```diff
@@ -314,6 +314,11 @@ <h3>Adding Affordances to Representations</h3>
       <code>comments</code> property in the following example is intended
       to be dereferenced.</p>
 
+    <p class="note">It is recommended to dereference resources that are
+      within API's domain. This may prevent possible issues with cross-site
+      scripting or obtaining meaningless resources. Still, there is no formal
+      prohibition of dereferencing <i>untyped</i> links, i.e. <i>rdf:seeAlso</i>.</p>
```
```suggestion
      prohibition of dereferencing resources link with well-know properties, e.g. <i>rdf:seeAlso</i>.</p>
```

spec/latest/core/index.html
```diff
@@ -314,6 +314,11 @@ <h3>Adding Affordances to Representations</h3>
       <code>comments</code> property in the following example is intended
       to be dereferenced.</p>
 
+    <p class="note">It is recommended to dereference resources that are
+      within API's domain. This may prevent possible issues with cross-site
+      scripting or obtaining meaningless resources. Still, there is no formal
```
tpluscode: I see what you did there but the word `meaningless` is a bit strong. 

```suggestion
      scripting or obtaining non-information resources. Still, there is no formal
```

Is that correct use of the term `non-information resource`? @asbjornu

tpluscode:
Also, please change to a specific title which corresponds the changes.

spec/latest/core/index.html
alien-mcl: For the record - I've replaced it as suggested, but I also have some doubts

spec/latest/core/index.html
```diff
@@ -314,6 +314,11 @@ <h3>Adding Affordances to Representations</h3>
       <code>comments</code> property in the following example is intended
       to be dereferenced.</p>
 
+    <p class="note">It is recommended to dereference resources that are
+      within API's domain. This may prevent possible issues with cross-site
```
asbjornu:
```suggestion
      within an API's domain. This may prevent possible issues with cross-site
```

spec/latest/core/index.html
asbjornu: I'm not sure what either "meaningless resources" or "non-information resources" is supposed to mean. Could you please elaborate?

spec/latest/core/index.html
asbjornu:
```suggestion
      prohibition of dereferencing resource links with well-known properties, e.g. <i>rdf:seeAlso</i>.</p>
```

spec/latest/core/index.html
tpluscode: Like I say, I'm not sure if this is correct use of the term non-information resource.

I understand this part to say "resources which have not meaning to the client". Or "resources which the client is unable to interpret".

Maybe we should just make this part more verbose but unambiguous?

alien-mcl: I agree - I'll try to update it ASAP.

alien-mcl: I've rephrased it so it's more verbose now.

spec/latest/core/index.html
```diff
@@ -314,6 +314,13 @@ <h3>Adding Affordances to Representations</h3>
       <code>comments</code> property in the following example is intended
       to be dereferenced.</p>
 
+    <p class="note">It is recommended to dereference resources that are
+      within an API's domain. This may prevent possible issues with cross-site
+      scripting or obtaining resources which might have no meaning
+      to the client or the client would be unable to interprete.
+      Still, there is no formal prohibition of dereferencing resources
+      link with well-know properties, e.g. <i>rdf:seeAlso</i>.</p>
```
asbjornu:
```suggestion
      links with well-known properties, e.g. <i>rdf:seeAlso</i>.</p>
```

spec/latest/core/index.html
```diff
@@ -314,6 +314,13 @@ <h3>Adding Affordances to Representations</h3>
       <code>comments</code> property in the following example is intended
       to be dereferenced.</p>
 
+    <p class="note">It is recommended to dereference resources that are
+      within an API's domain. This may prevent possible issues with cross-site
+      scripting or obtaining resources which might have no meaning
+      to the client or the client would be unable to interprete.
+      Still, there is no formal prohibition of dereferencing resources
```
asbjornu:
```suggestion
      Still, there is no formal prohibition of dereferencing resource
```

spec/latest/core/index.html
tpluscode: Actually, I think it should say "linked"

spec/latest/core/index.html
tpluscode: See above

alien-mcl: I think `dereferencing resources linked` fits better

Issue #32, opened 2014-02-04
Should hydra:returns and hydra:statusCodes be removed to avoid tight coupling?


lanthaler:
There are [concerns that `hydra:returns` and `hydra:statusCodes` promote tight coupling and automatic code generation](http://lists.w3.org/Archives/Public/public-hydra/2014Feb/0022.html). An option to avoid that, would be to simply remove them. The disadvantage of doing so would be that it becomes impossible to generate human-readable documentations similar to the ones of current Web APIs.

_On Tuesday, February 04, 2014 2:11 AM, Ryan J. McDonough wrote:_

> On Feb 3, 2014, at 3:19 PM, Markus Lanthaler wrote:
> 
> > I think the biggest problem is how this stuff is documented and the fact
> > that people won't read it :-) It's not a contract. It's a hint. Clients have
> > to interpret the response in any case. Some will be more tolerant, some will
> > break when they don't get back what they expected. That's similar to how
> > people often run into troubles when parts of a website get redesigned and
> > "nothing works anymore" because it looks different or they have to take
> > different paths.
> 
> What I was trying to get at with the profile link header is that if the data
> model is expressed up front and there’s sufficient documentation about what
> the types mean, a client can create a number of handlers that could react to
> different responses. If the model is good enough, the client could react
> better to responses that they didn’t expect at build time.
> 
> Some developers will read documentation and create code generators that the
> illiterate ones will use. It is here that I feel Hydra will get into trouble.
> 
> > I have troubles extracting something actionable from your mails. Would
> > removing returns/statusCodes address your concerns?
> 
> Absolutely!
> 
> > If so, what does it really change?
> 
> It'll force developers to look at at what’s coming back in the response
> headers rather than what’s defined in the Hyrda description. One is a hint and
> the other is fact (i.e. what the server is sending back). By removing returns
> and status, you are now forcing developers to look in the right place: the
> HTTP response headers.
> 
> I have development teams messing things up on a fairly frequent basic with
> WADL and Swagger (seeing a pattern here? :) ) due to the fact that they are
> expecting the server to return exactly what is specified in the descriptor and
> not taking into account they may have to deal with both a forward and reverse
> proxy in the mix.
> 
> It could also be the wording. If this is just a hint, then perhaps instead of
> hydra:returns, which sounds a bit more committed than say perhaps something
> more like hyrdra:intimation or perhaps even hydra:anticipatedResponseType?
> 
> I’m still not sold on status codes. For the most part, everyone is going to
[...]

PR #175, opened 2019-01-09
Should returns and statusCodes be removed to avoid tight coupling?

alien-mcl:
There was already a paragraph that mentioned a _volatile_ character of _ApiDocumentation_, which I've extended with direct references to _hydra:returns_ and _hydra:statusCodes_ hoping it will make the #32 resolved.

### DIFF
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -555,9 +555,12 @@
 
     <p>Keep in mind that operations specified in an <i>ApiDocumentation</i>
       may fail at runtime as either resources or the <i>ApiDocumentation</i>
-      itself have changed since they have been retrieved. A simple strategy
-      to try to recover from such an error is to reload the
-      <i>ApiDocumentation</i>.</p>
+      itself have changed since they have been retrieved. Also operation details
+      like <i>hydra:returns</i> or <i>hydra:statusCodes</i> may vary at runtime,
+      thus these interpreted at runtime. A simple strategy to try to recover from
+      such a situation is to reload the <i>ApiDocumentation</i> and redo all
+      operations that were based on the <i>ApiDocumentation</i> (or at least those
+      that lead to the current failure).</p>
 
     <p class="issue">Describe the various properties of an operation.</p>
   </section>
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -556,11 +556,14 @@
     <p>Keep in mind that operations specified in an <i>ApiDocumentation</i>
       may fail at runtime as either resources or the <i>ApiDocumentation</i>
       itself have changed since they have been retrieved. Also operation details
-      like <i>hydra:returns</i> or <i>hydra:possibleStatus</i> may vary at runtime,
-      thus these interpreted at runtime. A simple strategy to try to recover from
-      such a situation is to reload the <i>ApiDocumentation</i> and redo all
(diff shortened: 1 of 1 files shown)


spec/latest/core/index.html
```diff
@@ -875,7 +878,7 @@ <h3>Description of HTTP Status Codes and Errors</h3>
       {
         "@context": "http://www.w3.org/ns/hydra/context.jsonld",
         "@type": "****Status****",
-        ****"statusCode": 429****,
+        ****"possibleStatus": 429****,
```
tpluscode: Please remind me when that changed

spec/latest/core/index.html
```diff
@@ -555,9 +555,12 @@ <h3>Documenting a Web API</h3>
...
       may fail at runtime as either resources or the <i>ApiDocumentation</i>
-      itself have changed since they have been retrieved. A simple strategy
-      to try to recover from such an error is to reload the
-      <i>ApiDocumentation</i>.</p>
+      itself have changed since they have been retrieved. Also operation details
+      like <i>hydra:returns</i> or <i>hydra:possibleStatus</i> may vary at runtime,
+      thus these interpreted at runtime. A simple strategy to try to recover from
+      such a situation is to reload the <i>ApiDocumentation</i> and redo all
+      operations that were based on the <i>ApiDocumentation</i> (or at least those
+      that lead to the current failure).</p>
```
tpluscode: I don't see how this is relevant. Furthermore, I think this is incorrect and misleading.

Why would the client redo operations which succeeded prior to reloading the API Documentation?

>  operations that were based on the ApiDocumentation

The language here is also weird. How is an operation based on the ApiDocumentation? This terminology is not used anywhere else AFAICT.

spec/latest/core/index.html
```diff
@@ -555,9 +555,12 @@ <h3>Documenting a Web API</h3>
 
     <p>Keep in mind that operations specified in an <i>ApiDocumentation</i>
       may fail at runtime as either resources or the <i>ApiDocumentation</i>
-      itself have changed since they have been retrieved. A simple strategy
-      to try to recover from such an error is to reload the
-      <i>ApiDocumentation</i>.</p>
+      itself have changed since they have been retrieved. Also operation details
+      like <i>hydra:returns</i> or <i>hydra:possibleStatus</i> may vary at runtime,
+      thus these interpreted at runtime. A simple strategy to try to recover from
```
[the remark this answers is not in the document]
alien-mcl: I think `MUST` is to strong - I'd prefer `SHOULD`.
When the payload is a text, indeed it's quite easy to determine correctness of what we've received, but in case of binary payload (I know - Hydra does not support these yet), it might be tricky to figure it out whether the blob is an AVI movie with DivX encoding we've expected, or an AVI movie, but in h264 encoding. HTTP Content-Type header won't help you much - for both it'll be _video/avi_.

spec/latest/core/index.html
alien-mcl: #27

spec/latest/core/index.html
alien-mcl: I've reworded _operation_ with _pre-computations_. In general, I can imagine a form is generated from a given ApiDocumentation, but the received resource does not match the ApiDocumentation. Resolution would require the application to reload ApiDocumentation, regenerate the form and apply the received resource again.

I've also added an example involving error message shown to the user with option to return to e.g. home screen.

Issue #48, opened 2014-06-15
Add domainIncludes/rangeIncludes statements


lanthaler:
**_[Proposed by Jindřich Mynarz](http://lists.w3.org/Archives/Public/public-hydra/2014Jun/0069.html):**_

One of the key aids that helped me to understand Hydra is its [diagram](http://www.hydra-cg.com/spec/latest/core/vocabulary.png), which provides a good overview of how Hydra's vocabulary terms are related. However, what I find confusing is that some of the relations depicted in the diagram aren't present in Hydra's description in RDF. In particular, I'm missing `rdfs:domain` (or `schema:domainIncludes`) axioms, which would help me (and, presumably, other developers with semantic web background) to orient in the vocabulary. Having `rdfs:domain` explicitly set for properties would help to recognize which classes are the properties compatible with; even if this is to be treated with open world assumption as just a hint. 

I've scanned through the RDF description of the Hydra Core Vocabulary and found the following properties, which could be enriched with `rdfs:domain` or `schema:domainIncludes`, which would likely be a better fit in this case. For each property I tried to guess, mostly using the [Hydra's vocabulary diagram](http://www.hydra-cg.com/spec/latest/core/vocabulary.png), what are its compatible classes.
- `hydra:apiDocumentation` - I assume that any `hydra:Resource` is compatible with this property.
- `hydra:description` - `hydra:description schema:domainIncludes hydra:ApiDocumentation, hydra:StatusCodeDescription .`
- `hydra:freetextQuery` - `hydra:Resource`? 
- `hydra:property` - `schema:domainIncludes hydra:SupportedProperty, hydra:IriTemplateMapping .`
- `hydra:required` - `schema:domainIncludes hydra:SupportedProperty, hydra:IriTemplateMapping .`
- `hydra:search` - `hydra:Resource`?
- `hydra:statusCodes` - `schema:domainIncludes hydra:ApiDocumentation, hydra:Operation .`
- `hydra:supportedOperation` - `schema:domainIncludes hydra:Class, hydra:Link .`
- `hydra:title` - `schema:domainIncludes hydra:ApiDocumentation, hydra:StatusCodeDescription .`

Is this a more or less correct understanding? I think having these relationships explicitly documented in the vocabulary (and not only in the vocabulary's diagram) would make its intended uses clearer.

alien-mcl:
Closed with PR #176

PR #176, opened 2019-01-09
Add domainIncludes/rangeIncludes statements

alien-mcl:
I've added some _domain_ predicates. There are other terms that would need this kind of specification, but they are applicable to more than type of resource. I hesitated to use both _schema:domainIncludes_ or _owl:unionOf_ based constructs. I didn't see any such a logic process that utilized _schem_ predicate and using _owl_ would bring mayhem to the spec.
Feel free to deliberate more.

### DIFF
--- a/spec/latest/core/core.jsonld
+++ b/spec/latest/core/core.jsonld
@@ -113,6 +113,7 @@
       "label": "apiDocumentation",
       "comment": "A link to the API documentation",
       "range": "hydra:ApiDocumentation",
+      "domain": "hydra:Resource",
       "vs:term_status": "testing"
     },
     {
@@ -407,6 +408,7 @@
       "label": "search",
       "comment": "A IRI template that can be used to query a collection.",
       "range": "hydra:IriTemplate",
+      "domain": "hydra:Resource",
       "vs:term_status": "testing"
     },
     {
@@ -415,6 +417,7 @@
       "label": "freetext query",
       "comment": "A property representing a freetext query.",
       "range": "xsd:string",
+      "domain": "hydra:Resource",
       "vs:term_status": "testing"
     },
     {
--- a/spec/latest/core/core.jsonld
+++ b/spec/latest/core/core.jsonld
@@ -8,6 +8,7 @@
     "vs": "http://www.w3.org/2003/06/sw-vocab-status/ns#",
     "dc": "http://purl.org/dc/terms/",
     "cc": "http://creativecommons.org/ns#",
+    "schema": "http://schema.org/",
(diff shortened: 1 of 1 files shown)


tpluscode:
Issue #48 proposes a number of statements using `schema:domainIncludes`. I don't think it's necessary to use with OWL but rather as simple bag

```
{
  "@id": "title",
  "schema:domainIncludes": [
    "ApiDocumentation",
    "StatusCodeDescription"
  ]
```

Let's evaluate the proposed statements and decide if they are indeed correct.

tpluscode:
`rdfs:domain` has strong semantics, essentially stating that the object is of given type, right?

Maybe we should just use `domainIncludes` everywhere, intending it as a human-hint and not adding RDF semantics?

@RubenVerborgh

vddesai1871:
Related to predicates missing domain, how come `hydra:required` doesn't have domain specified but `hydra:readable` and `hydra:writeable` have? I suppose it wasn't intentional. 

> There are other terms that would need this kind of specification, but they are applicable to more than type of resource.

In that case if they have some common superclass then we can use that as `domain`.

alien-mcl:
>Maybe we should just use domainIncludes everywhere,
>intending it as a human-hint and not adding RDF semantics?

Retracting current `rdfs` statements would be a breaking change - let's stay with adding `schema:domainIncludes`

alien-mcl:
In the last commit I've also extended list of `domainIncludes` statements for `title` and `description` to cover more `hydra` types. It was not on the _image_ on the spec page, but I think having possibility to come with a custom description i.e. on `SupportedProperty` would be nice.

tpluscode:
Oh, nice, Is this auto-generated?

alien-mcl:
Oh, no. Hand-crafted. I wanted to have an actual view of Hydra Core Vocabulary before attempt to add some new features

tpluscode:
Any chance to script its generation? We could have it created automatically

vddesai1871:
```
{
      "@id": "hydra:collection",
      "@type": "hydra:Link",
      "label": "collection",
      "comment": "Collections somehow related to this resource.",
      "domain": "hydra:Collection",
      "vs:term_status": "testing"
},
```

Shouldn't it have `hydra:Collection` as range? and `hydra:Class` as domain..( as the comment says "Collections somehow related to this resource." )?

vddesai1871:
And if I am correct `hydra:template` will have `IriTemplate` as domain, I am currently working on it at hydrus (hydra-ecosystem). If you haven't got time I can create a PR to your branch adding these changes.

PR #177, opened 2019-01-09
Added an example of Api response -> ApiDocumentation flow

### LINKED ISSUE #69: Describe entry point -> ApiDocumentation flow
**_[Martijn Faassen raised this](http://lists.w3.org/Archives/Public/public-hydra/2014Sep/0025.html):**_

> I'd like to see the use case described where the entry point comes
> first, it points to the Hydra docs, and the Hydra docs doesn't have to
> specify an entry point. And then secondarily describe the use case you
> have now where the entry point is embedded in the Hydra content

alien-mcl:
As in the title, I've added an example of Api response -> ApiDocumentation flow, which I hope will close #69

### DIFF
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -563,7 +563,7 @@
   </section>
 
   <section>
-    <h3>Discovering a Hydra-powered Web API</h3>
+    <a name="discovering-api"></a><h3>Discovering a Hydra-powered Web API</h3>
 
     <p>The first step when trying to access a Web API is to find an entry
       point. Typically, this is done by looking for documentation on the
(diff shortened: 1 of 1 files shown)


spec/latest/core/index.html
```diff
@@ -630,6 +630,40 @@ <h3>Discovering a Hydra-powered Web API</h3>
       using HTTP Link Headers is typically not necessary as the concepts
       used in responses from the API will dereference to their
       documentation.</p>
+
+    <p>Another case would involve an <i>ApiDocumentation</i> discover from
+      another Api response. There is no restriction on emitting an HTTP
```
```suggestion
      a bookmarked resource's representation. There is no restriction on emitting an HTTP
```

spec/latest/core/index.html
```diff
@@ -630,6 +630,40 @@ <h3>Discovering a Hydra-powered Web API</h3>
       using HTTP Link Headers is typically not necessary as the concepts
       used in responses from the API will dereference to their
       documentation.</p>
+
+    <p>Another case would involve an <i>ApiDocumentation</i> discover from
```
```suggestion
    <p>In another scenario the <i>ApiDocumentation</i> would be discovered from
```

spec/latest/core/index.html
tpluscode: I think that the sentence about restricting the `Link` header is redundant. In other words, is there anything in the spec suggesting otherwise?

Actually, I'd got one step further and actually state that a Hydra API SHOULD emit the api doc link every time

spec/latest/core/index.html
```diff
@@ -630,6 +630,40 @@ <h3>Discovering a Hydra-powered Web API</h3>
...
+      GET /api/items HTTP/1.1
+      Host: api.example.com
+      Accept: application/ld+json
+
+      ====================================
+
+      HTTP/1.1 200 OK
+      ...
+      Content-Type: application/ld+json
+      ****Link: <http://api.example.com/doc/>; rel="http://www.w3.org/ns/hydra/core#apiDocumentation"****
```
tpluscode: Do asterisks here somehow affect formatting?

spec/latest/core/index.html
```diff
@@ -630,6 +630,40 @@ <h3>Discovering a Hydra-powered Web API</h3>
...
+        "@id": "http://api.example.com/api/items",
+        "title": "Items collection",
+        ...
+      }
+      -->
+    </pre>
+
+    <p class="note">It is possible to use a fallback behavior in case an
+      Api response does not present an HTTP <i>Link</i> header. In such a situation,
+      client may use <i>Host</i> HTTP header value that was used to receive a response
```
tpluscode: I don't follow. Where is the `Host` header behaviour described?

spec/latest/core/index.html
alien-mcl: It gets greenish in the final spec

spec/latest/core/index.html
alien-mcl: Ok, maybe I should rephrase it. In general, I wanted to enable a fallback behavior in a situation when call to `http://some.url/api/items` does not provide an _ApiDocumentation_ link, client could call `http://some.url/` to obtain it (where `some.url` is actually a `Host`)

tpluscode: I don't think we should have fallback defined that way. `Host` will be greatly unreliable. In most cases it may not even be intended for dereferencing.

Just the more reason to have ApiDoc mandatory for every response.

alien-mcl: I've softened the Link header requirement with SHOULD

Issue #178, opened 2019-01-11
Make hydra:error compatible with RFC 7807


asbjornu:
As discussed in #39, it would be useful if `hydra:error` responses were compatible with [RFC 7807](https://tools.ietf.org/html/rfc7807). Defining a `@context` for `application/problem+json` that defines `"instance": "@id"` should achieve what @RubenVerborgh is after, I believe.

tpluscode:
Yes, we'd need a `@context`, which would also be define the [other properties](https://tools.ietf.org/html/rfc7807#section-3.1) of the problem details object. 

Guess that for starters we could publish it under hydra's domain but how would you define RFC 7807 terms? I think I recall talk of similar effort with @dret. Isn't there already an attempt to create JSON-LD profile for problem+json? Or, in more general terms, an RDF vocabulary for the Problem Details spec?

The end result could be similar to the following

```json
{
    "@context": [
        "http://www.w3.org/ns/hydra/context.jsonld",
        "http://www.w3.org/ns/hydra/problem+json/context.jsonld",
        "https://example.com/errors.context"
    ],
    "@type": [
        "Error",
        "https://example.com/probs/out-of-credit"
    ],
    "type": "https://example.com/probs/out-of-credit",
    "title": "You do not have enough credit.",
    "detail": "Your current balance is 30, but that costs 50.",
    "instance": "/account/12345/msgs/abc",
    "balance": 30,
    "accounts": ["/account/12345", "/account/67890"]
   }
```

Some things to note:

1. Additional `@context` would be added to define API-specific error details (`balance`)
2. `type` is problematic and would have to be duplicated if one were to also use it as an RDF type. The RFC defines it as `string` and so it cannot be simply redefined as `@type`

dret:
On 2019-01-11 05:01, Tomasz Pluskiewicz wrote:
> Guess that for starters we could publish it under hydra's domain but how 
> would you define RFC 7807 terms? I think I recall talk of similar effort 
> with @dret <https://github.com/dret>. Isn't there already an attempt to 
> create JSON-LD profile for problem+json? Or, in more general terms, an 
> RDF vocabulary for the Problem Details spec?

i remember somebody being interested about this a while ago and i 
offered to help with that effort if needed, but afaict, this never went 
anywhere.

tpluscode:
Ok, that is good. We have a clean slate.

[...]

PR #180, opened 2019-01-17
expand on @id with `Status`/`Error` response

### LINKED ISSUE #39: Document how "errors" can be given an identifier and be reused in responses
Errors or possible statuses in general can be given an identifier that can then be directly reused in responses. According to [RubenVerborgh this should be documented in the spec](http://lists.w3.org/Archives/Public/public-hydra/2014Mar/0004.html):

> > > But doesn't really help anyway. Clients won't do anything with the
> > > description.
> > 
> > That might not be entirely true as you could give that specific status an
> > identifier and then return it as is as response:
> > 
> > ```
> >   {
> >     "@context": "http://www.w3.org/ns/hydra/context.jsonld",
> >     "@id": "http://example.com/api/errors/180984",
> >     "@type": "Status",
> >     "title": "Too Many Requests",
> >     "description": "A maximum of 500 requests per hour and user is allowed."
> >    }
> > ```
> 
> I like that very much.
> So much that I would actually recommend in the spec
> that the status is given an identifier, that it's not just a blank node.
> That way, it can be annotated easily later on.

tpluscode:
fixes #39 

Once we have something to show for #178, we could expand this section to further suggest the use of problem+json. At this point I don't think there is enough to it to include here.

### DIFF
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -898,6 +898,10 @@
       Hydra defines just one such subclass, namely the <i>Error</i> class.
       This provides an extensible framework to communicate error details to
       a client.</p>
+    
+    <p>Furthermore, a <i>Status</i> or <i>Error</i> returned by the server can also
+      be given an identifier. When dereferenced, the <i>Error</i> resource can provider
+      more detailed information or possible ways to resolve the problem, if applicable.<p/>
 
     <pre class="example nohighlight" data-transform="updateExample"
          title="Using Hydra's Error class to describe errors">
@@ -908,6 +912,7 @@
       {
         "@context": "http://www.w3.org/ns/hydra/context.jsonld",
         "@type": "****Error****",
+        "@id": "http://api.example.com/error-details/1234",
         "title": "An error occurred",
         "description": "Typically, a specialization of this class is used in practice.",
         ...
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -900,7 +900,7 @@
       a client.</p>
     
     <p>Furthermore, a <i>Status</i> or <i>Error</i> returned by the server can also
-      be given an identifier. When dereferenced, the <i>Error</i> resource can provider
+      be given an identifier. When dereferenced, the <i>Error</i> resource can provide
       more detailed information or possible ways to resolve the problem, if applicable.<p/>
 
     <pre class="example nohighlight" data-transform="updateExample"


spec/latest/core/index.html
```diff
@@ -898,6 +898,10 @@ <h3>Description of HTTP Status Codes and Errors</h3>
       Hydra defines just one such subclass, namely the <i>Error</i> class.
       This provides an extensible framework to communicate error details to
       a client.</p>
+    
+    <p>Furthermore, a <i>Status</i> or <i>Error</i> returned by the server can also
+      be given an identifier. When dereferenced, the <i>Error</i> resource can provider
```
alien-mcl:
```suggestion
      be given an identifier. When dereferenced, the <i>Error</i> resource can provide
```

spec/latest/core/index.html
```diff
@@ -898,6 +898,10 @@ <h3>Description of HTTP Status Codes and Errors</h3>
...
+      more detailed information or possible ways to resolve the problem, if applicable.<p/>
```
alien-mcl: How an `Error` can do that?
There is indeed an example of a `application/problem+json` with `invalid-params` property added on the [RFC 7807](https://tools.ietf.org/html/rfc7807), but it's not enlisted as a valid member, neither in that nor in hydra's spec.

spec/latest/core/index.html
```diff
@@ -908,6 +912,7 @@ <h3>Description of HTTP Status Codes and Errors</h3>
       {
         "@context": "http://www.w3.org/ns/hydra/context.jsonld",
```
alien-mcl: I'd go with problem specific context

spec/latest/core/index.html
tpluscode: Not problem+json itself, but the linked resource. The RFC defines an `instance` property:

```json
{
    "type": "https://example.com/probs/out-of-credit",
    "instance": "/account/12345/msgs/abc"
}
```

That, when dereferenced, can include specific Hydra operations.

spec/latest/core/index.html
tpluscode: I agree but we're not there yet. IMO this has to wait until there is more work done in #178

alien-mcl: Hmm. Is there any chance to expand this PR with additional context? Or shall we have a separate PR for that later

tpluscode: I think we'll add it later. Currently there is nothing we'd put in such context

tpluscode: > Well, it would have details of how RFC 7807 problem would be translated to hydra.

Like I said, we'd have to work on #178 first. The result would be that context.
Otherwise, this snippet has no relation with problem+json.

> Then, we could have it linked to hydra's main context.

Out of scope of this PR, but I think I'd have ot the opposite way

alien-mcl:
Can we proceed with this one somehow?

tpluscode:
Guess that if you approved and no one else disapproves, then you're at liberty to merge

tpluscode:
...once we agree during the call?

Issue #102, opened 2015-10-29
Define client-initiated pagination


lanthaler:
In issues #41 and #42 we defined how to represent collections and paginated collections. It became clear that people want a mechanism to tell a clients how to control pagination and how jump directly to specific pages/views.

Requirements/desired features are collected at https://www.w3.org/community/hydra/wiki/Client-initiated_pagination

tpluscode:
AFAICT, the current recommendation is to use `search` property to allow the client to narrow down/navigate through a collection.

In te use cases we have:

```
{
    "search": {
        "@type": "IriTemplate",
        "template": "http://example.com/api/events{?search}",
        "variableRepresentation": "BasicRepresentation",
        "mapping": [
            {
                "@type": "IriTemplateMapping",
                "variable": "search",
                "property": "freetextQuery",
                "required": true
            }
        ]
    }
}
```

I think we can simply provide some well-known properties like `freetextQuery` to support common scenario while not being too invasive.

tpluscode:
To jump to a specific page, the server could instead advertise a template like `http://example.com/api/events{?search,page}`.

A simple start would be to add a mapping. 

```diff
{
  "mapping": [
+    "@type": "IriTemplateMapping",
+    "variable": "page"
+    "property": "page"
+    "required": false
  ]
}
```

Not too happy about `page` but any other name I can think of suggests a data type. Typically the pages are numbers, but we cannot enforce that on APIs.

tpluscode:
Another common design choice could be `skip`/`take` approach:

```diff
{
-  "template": "http://example.com/api/events{?search}",
+  "template": "http://example.com/api/events{?search,limit,offset}",
  "mapping": [
+  {
+    "@type": "IriTemplateMapping",
+    "variable": "limit",
+    "property": "take",
+    "required": false
+  },
+  {
+    "@type": "IriTemplateMapping",
+    "variable": "offset",
+    "property": "skip",
+    "required": false
+  }
  ]
}
```

angelo-v:
How do we associate the IriTemplate with a PartialCollectionView?

tpluscode:
@angelo-v the template is not directly related to the Collection(View). 

Templates can be used for links (`search` or any other) to let the client fill in the blanks.

angelo-v:
But in this case it would be related to a Collection via hydra:search?

angelo-v:
[...]

PR #182, opened 2019-01-30
Vocabulary extensions 

### LINKED ISSUE #22: Allow returns/expects to be expressed in terms of a media type; what about multiple types?
This is, e.g., necessary for things such as images. It is also necessary to define how multiple types in `expects`/`returns` should be interpreted. Do requests/responses need to be instances of _all_ classes or just of at least one?

### LINKED ISSUE #99: Add support for describing headers accepted in supported operations
Some web applications (e.g., LDP apps) make use of HTTP headers such as Slug, Prefer for certain operations. It would help the client if this support can be explicitly described in the API documentation.  Further, it may be useful if the server can describe the range of the header, i.e., the values the server is willing to accept using  a template, a regex pattern, or an enumeration of values it will help the client to formulate a valid request. 

The initial discussion on Hydra mailing list is [here](https://lists.w3.org/Archives/Public/public-hydra/2015Mar/0013.html)

### LINKED ISSUE #160: Expressing actions (#154) within ApiDocumentation
On today's call we discussed that the premise of PR #154 is to allow, as the title states, performing requests on a different target. That is, as opposed to `hydra:Operation` which always assumes that the target will be the object of a `?s hydra:operation ?o` triple.

Taking advantage of `supportedProperty` and `supportedOperation` combined should make it also possible.

Let's have a car resource, linked to a steering wheel:

``` json
{
  "@type": [ "ex:Car" ],
  "@id": "/the/car"
  "ex:steeringWheel": {
    "@id": "/the/wheel"
  }
}
```

We can currently define the ApiDocumentation so that any occurrence of `ex:steeringWheel` implies that it's subject will support given operations:

``` json
{
  "supportedClass": {
    "@id": "ex:Car",
    "supportedProperty": {
      "@id": "ex:steeringWheel",
      "supportedOperation": [ {
        "@type": "ex:TurnLeftOperation"
      }, {
        "@type": "ex:TurnRightOperation"
      }
    }
  }
}
```

This way any subject of `?car ex:steeringWheel ?subject` will be a assumed a valid for the turn operations, regardless of it's precise class. I think it should be also allowed to have an `IriTemplate` in place of  `/the/wheel`

``` json
{
  "@type": [ "ex:Car" ],
  "@id": "/the/car"
  "ex:steeringWheel": {
    "@type": "IriTemplate"
  }
}
```

@lanthaler please confirm I'm not inventing my own Hydra here.
@elf-pavlik how would similar be expressed with ApiDocumentation and the actions being discussed in #154

### LINKED ISSUE #167: API documentation limitations
Currently it is not possible to mention IRI templates within the API documentation.
There is a PR #164 that tries to describe as much as possible of an example API which hit a wall of that limitation.
There are also other limitations encountered in that PR mentioned that would need to be resolved in order to win more spec users.

alien-mcl:
Extended vocabulary with these features:
- enable operations for non-RDF resources (resolves #22)
- enable operations for headers (resolves #99)
- enable operations with explicit operation target (API documentation) (resolves #160, obsoletes #154 )
- enable strongly typed collections (API documentation) (resolves #167)
- enable client-side initiated pagination (resoles #102)

In general, these _imperfect_ attempt to extend the Hydra Core Vocabulary (HCV) should enable the it for some of the uses cases that appeared on various occasions. Feel free to deliberate more, but I hope these will find their way to the final specification.

I tried to keep backward compatibility so current implementations using the vocabulary in it's current state won't get broken. Unfortunately, there are some minor changes to some of the elements.

I've also added an updated version of the [vocabulary.png](https://raw.githubusercontent.com/HydraCG/Specifications/1f3adabb25ede5635ba7df236d75f3f4cb14928a/spec/latest/core/vocabulary.png) diagram (including source code).

I'll try to provide a proper PR for [Heracles.ts](/HydraCG/Heracles.ts) reference client covering these features ASAP.


spec/latest/core/core.jsonld
```diff
@@ -505,6 +518,112 @@
...
+    {
+      "@id": "hydra:HeaderTemplate",
+      "@type": "rdfs:Class",
+      "subClassOf": "hydra:Template",
+      "label": "Header template",
+      "comment": "Specification of the header template with variables to be replaced with actual values.",
+      "vs:term_status": "testing"
+    },
+    {
+      "@id": "hydra:name",
```
angelo-v: hydra:name sounds very generic, as if you could name anything. Only the comment states, it is used to name a header. I would prefer hydra:headerName

alien-mcl: Good point. Here you go!

alien-mcl:
I've just pushed [PR](/HydraCG/Heracles.ts/pull/52) to [Heracles.ts](/HydraCG/Heracles.ts) with support for these features.

spec/latest/core/core.jsonld
```diff
@@ -505,6 +518,112 @@
...
+      "@type": "rdfs:Class",
+      "label": "Template",
+      "comment": "Some abstract template.",
+      "vs:term_status": "testing"
+    },
+    {
+      "@id": "hydra:Header",
+      "@type": "rdfs:DataType",
+      "label": "Header",
+      "comment": "Specification of the header, including name and value(s).",
```
tpluscode: I wonder, what is the current stand on Hydra vs HTTP?

I mean, adding headers to the core signifies a close relation with the protocol. Did we ever decide to keep Hydra somewhat agnostic in that regard?

@RubenVerborgh @lanthaler

spec/latest/core/core.jsonld
```diff
@@ -505,6 +518,112 @@
...
+      "vs:term_status": "testing"
+    },
+    {
+      "@id": "hydra:CollectionSpecification",
+      "@type": "rdfs:Class",
+      "subClassOf": "hydra:Collection",
+      "label": "Collection specification",
+      "comment": "Describes a collection returned by the operation.",
+      "vs:term_status": "testing"
+    },
```
tpluscode: Where does that come from? How would this come up against the "manages block"? Sounds similar

spec/latest/core/core.jsonld
```diff
@@ -505,6 +518,112 @@
...
+      "range": "hydra:MediaType",
+      "vs:term_status": "testing"
+    },
+    {
+      "@id": "hydra:MediaType",
+      "@type": "rdfs:DataType",
+      "label": "Media type",
+      "comment": "An RFC 6838 compliant media type specification.",
+      "vs:term_status": "testing"
+    },
```
tpluscode: The media type stuff is cool in principle but how does it relate to the other changes within this PR?

alien-mcl:
> I strongly disapprove of this pull request!

Ok. Indeed this is an imperfect attempt.

> looks like there are several unrelated changes

Indeed - there are several topics onboard
> none of the proposed changes have a related (or mentioned) PR and
> have not gone through some analysis beforehand

I partially disagree - operations with explicit targets were discussed quite extensively; other thing it didn't end up with any specific conclusion

> it literally came out of the blue

I disagree - most of topics touched here were either discussed or at least mentioned; I also did advertised this attempt in various occasions

> Instead of haphazardly submitting PRs we must follow a more structured
> workflow and follow some ground rules as closely as possible

What kind of structured workflow? I agree that we didn't provide any detailed roadmap for future work, but GH is full of unresolved issues that needs to be addressed and should be treated as that roadmap. I remember how we did work in last several years and loo were we are now. 

> each PR should address a specific issue or at least related issues

I can cherry-pick each implemented feature as a separate PR if that will work for you


> each issue should be first discussed, deemed important and have a rough idea of the solution

As I already said - some of the features were discussed - some of them indeed may need some discussion. This is why that PR came anyway. I don't want to raise new issues to address other issues - it's pointless. 

> PR with new features should not only introduce vocabulary elements
> but also extend upon the human-readable part of the spec.

I may agree with this, but this would imply enormous effort on me still having a large risk of a complete failure. Once I confirm that these changes are going in the right direction, I may consider spec changes

I could go one step further and suggest that the prose should even come slightly before vocabulary. That way it could be easier to understand the intent and gauge the effect of said changes. On the other hand a discussion under the related issue would serve that purpose too.

spec/latest/core/core.jsonld
```diff
@@ -59,6 +59,19 @@
     "mapping": "hydra:mapping",
     "IriTemplateMapping": "hydra:IriTemplateMapping",
     "variable": "hydra:variable",
+    "skip": { "@id": "hydra:skip", "@type": "xsd:integer" },
+    "take": { "@id": "hydra:take", "@type": "xsd:integer" },
```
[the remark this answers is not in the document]
alien-mcl: If it is a desired feature, it should be provided in the core. I'm opened for alternative approaches though. These are just terms that are commonly used in other querying languages like SQL (TOP, LIMIT), LINQ (Skip, Take), etc.

spec/latest/core/core.jsonld
```diff
@@ -505,6 +518,112 @@
       "domain": "hydra:IriTemplateMapping",
       "range": "xsd:string",
       "vs:term_status": "testing"
+    },
+    {
+      "@id": "hydra:target",
+      "@type": "rdf:Property",
+      "label": "invocation target",
+      "comment": "Explicit target of the invocation of the operation.",
```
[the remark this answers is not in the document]
alien-mcl: Yep, as it is stated in the PR's description at the very beginning

spec/latest/core/core.jsonld
```diff
@@ -505,6 +518,112 @@
...
+    },
+    {
+      "@id": "hydra:returnsHeader",
+      "@type": "rdf:Property",
+      "label": "returns header",
+      "comment": "Name of the header returned by the operation.",
+      "domain": "hydra:Operation",
+      "range": "xsd:string",
+      "vs:term_status": "testing"
+    },
```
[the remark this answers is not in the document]
alien-mcl: I actually tried to be consistent with `returns`/`expects`. Indeed, for returned header `possible` may work. For _expected_ it won't

spec/latest/core/core.jsonld
```diff
@@ -505,6 +518,112 @@
...
+    {
+      "@id": "hydra:expectsHeader",
+      "@type": "rdf:Property",
+      "label": "expects header",
+      "comment": "Specification of the header expected by the operation.",
+      "domain": "hydra:Operation",
+      "rangeIncludes": ["hydra:Header", "hydra:HeaderTemplate"],
+      "vs:term_status": "testing"
+    },
```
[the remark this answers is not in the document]
alien-mcl: While I believe both returned and expected headers should be connected to each other conceptually, this name is OK for me. For `returned` it should be then `responseHeader`

spec/latest/core/core.jsonld
```diff
@@ -505,6 +518,112 @@
...
+    {
+      "@id": "hydra:Template",
+      "@type": "rdfs:Class",
+      "label": "Template",
+      "comment": "Some abstract template.",
+      "vs:term_status": "testing"
+    },
```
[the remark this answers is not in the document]
alien-mcl: I was for backward compatibility - I didn't want to introduce another term for template in headers, thus I had to revamp the domain of the existing one. I also didn't want to replace `domain` with other predicate, thus this is why the abstract

spec/latest/core/core.jsonld
```diff
@@ -505,6 +518,112 @@
...
+    {
+      "@id": "hydra:HeaderTemplate",
+      "@type": "rdfs:Class",
+      "subClassOf": "hydra:Template",
+      "label": "Header template",
+      "comment": "Specification of the header template with variables to be replaced with actual values.",
+      "vs:term_status": "testing"
+    },
```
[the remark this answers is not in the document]
alien-mcl: I was just wondering i.e. about an `Authorization` header where you mark the authorization scheme and the secret following it. Maybe other vendor-specific headers would require this.
In general, this is indeed a feature covering exotic scenarios.

spec/latest/core/core.jsonld
alien-mcl: There are `possibleStatus` and `method` - these are already quite strictly bound to the protocol. Also headers are not HTTP specific - other protocols may have similar notion. SOAP is a bad example here (as it can be over HTTP), but the envelope has a notion of the header as well.

spec/latest/core/core.jsonld
alien-mcl: Ther was an issue on denoting on the API what the collection describes. This was an attempt of enabling that part of the spec with such a description.

spec/latest/core/core.jsonld
alien-mcl: I wanted to enable hydra to expect/return resources other than RDF. One of the approach of describing a resource would be to use the media type, i.e. for JPG or DOCX file that would be expected by the API

spec/latest/core/core.jsonld
tpluscode: Point taken

tpluscode:
Without going into detailed discussions here I think this PR should be split into 3 PRs, which address the three linked issues it aims to resolve. The other two bullets should (non-RDF payloads and headers) should be first discussed as issues.

Splitting into smaller chunks will allow for focused discussions and prevent one change block an unrelated one from being merged.

On a technical note, it's best to avoid submitting PR from downstream `master` branch because such can cause conflicts once it's merged.

alien-mcl:
Yep - I've just noticed _master_ branch. One more reason to split it. As for new features - I'll try to dig through GH to find related issues. I'm pretty sure headers were mentioned somewhere.

alien-mcl:
I've divided it into separate PRs (#183, #184, #185 and #186) - this PR is obsoleted by those.

Issue #152, opened 2017-11-20
Usage of IriTemplate


tpluscode:
I've started implementing `IriTemplate` in Alcaeus, the old Heracles.

My API is simple:

``` js
const iriTemplate = new IriTemplate(); // part of representation/docs
const variables = {};

const expanded = iriTemplate.expand(variables);
```

Taking the example from specs:

``` json
{
  "@context": "http://www.w3.org/ns/hydra/context.jsonld",
  "@type": "IriTemplate",
  "template": "http://api.example.com/issues{?q}",
  "variableRepresentation": "BasicRepresentation",
  "mapping": [
    {
      "@type": "IriTemplateMapping",
      "variable": "q",
      "property": "hydra:freetextQuery",
      "required": true
    }
  ]
}
```

My question is about the actual contents of the `variables` dictionary.

I understand it should use `property` URIs as keys, which are they placed in the template variables?

``` json
{
  "http://www.w3.org/ns/hydra/core#freetextQuery": "search criteria"
}
```

To produce `http://api.example.com/issues?q=search%20criteria`?

alien-mcl:
I'd put my bet on `variable` instead.
```json
{
  "q": "search criteria"
}
```

This won't get non-RDF developers scared and seams cleaner. This also enables you to use existing libraries that handle URI templating. Having the mapping details you can translate predicates to variables.

tpluscode:
Yes, I'd though similar until today but see [this comment](https://github.com/HydraCG/Specifications/issues/30#issuecomment-60583053) under #30.

@lanthaler's examples look like the values from JSON-LD. It made me thing that maybe I was wrong after all.

I was also thinking that maybe both should be supported. So that I'd look for either `http://www.w3.org/ns/hydra/core#freetextQuery` or `q`.

alien-mcl:
Well, being prompt for both is still doable. I'd be careful with using raw JSON-LD structures. Some strongly typed API would be better, i.e.
```typescript
new IriTemplateBuilder()
    .withVariable("q").setTo("search criteria")
    .withVariable("http://some.uri/).setTo("text")
    .withVariable("f").setTo("me", "en")
    .asObject();
````

lanthaler:
Both are fine. Eventually I'd prefer to lean on the property URLs but for that to make sense, we need to specify in much more detail what they actually mean and what their relationship to the operation/target is.

Issue #82, opened 2015-01-05
Add support for allowed literals and allowed individuals


dschulten:
Currently there is no way in hydra to express a fixed number of possible values.
Proposal:
Have two new properties in the domain of `hydra:SupportedProperty`:

```
hydra:allowedIndividuals - allowed URI values for a supported property
hydra:allowedLiterals - allowed literal values for a supported property
```
The :allowedIndividuals is for allowed values which are URIs, it should be used to apply @type:@vocab in the context (is it possible to achieve the same effect as @type:@vocab directly in the hydra vocab?)

```
	"@context": {
		"@vocab": "http://schema.org/",
		"hydra": "http://www.w3.org/ns/hydra/core#",
      "hydra:allowedIndividuals": {
        "@type": "@vocab"
      }
```

OTOH, allowedLiterals is for scalar values like strings or numbers.

See http://tinyurl.com/ok6yafq for a real-life JSON-LD playground example of allowedIndividuals vs. allowedLiterals.
At http://tinyurl.com/nc3qfk3 there is another example which demonstrates how a custom allowed individual such as EVENT_CANCELLED would be mapped to schema:EventCancelled.

RubenVerborgh:
Isn't that part of ontological modelling, and thus part of the property?

PS What is the Hydra mailing list thread for this?

dschulten:
Hi Ruben,

I do not think the ontological model of a property covers it. On the one hand, sometimes not all possible values are predefined, but they may be extensible by individuals from other vocabs, see the usage of goodrelations enums from schema.org. On the other hand, while a value might occur on a class in general, it might not be suitable in a particular context. 
If you think rdfs:range, it is not about value constraints but inference, so it doesn't enumerate possible values at all (I've learned that much by now). Or do you have something else in mind?

I'll add the corresponding mailing list thread url later, it is there, albeit somewhat nested.

Cheers,
Dietrich

---- Ruben Verborgh schrieb ----

RubenVerborgh:
I [replied on-list](http://lists.w3.org/Archives/Public/public-hydra/2015Jan/0019.html).

alien-mcl:
I think @RubenVerborgh is right and allowed individuals in most cases are topic for modelling. I can see at least two options here:
- OWL - to heavy for me
- SHACL - it's more to describing CWA structures in OWA way (and I personally prefer it over OWL)

[...]

PR #184, opened 2019-02-11
Define client-initiated pagination

alien-mcl:
This is a derivative of a PR #182 that covers #102 :
- added skip/take predicates that can be used for IriTemplate variable mappings


spec/latest/core/core.jsonld
```diff
@@ -505,6 +507,22 @@
       "domain": "hydra:IriTemplateMapping",
       "range": "xsd:string",
       "vs:term_status": "testing"
+    },
+    {
+      "@id": "hydra:skip",
+      "@type": "rdf:Property",
+      "label": "skip",
+      "comment": "Instructs to skip N elements of the set.",
+      "range": "xsd:integer",
```
angelo-v: What's the domain of those properties?

tpluscode:
I remember how we were discussing this. `skip`/`take` is but just one possibility. The other common would be explicit page number.

I think we can address both by extending the IRI Template description. I'll continue on #102

spec/latest/core/core.jsonld
angelo-v: shall we add "domain": "hydra:IriTemplateMapping"?

alien-mcl: No domain is specified, enabling any kind of resource to have that property. It's purpose it to have a property that could be used indeed in IriTemplateMapping's _property_ relation. Domain of `hydra:IriTemplateMapping` would be incorrect here.

alien-mcl:
I've decided on using skip/take (alternatives are offset/limit) as more generic than pageIndex/itemsPerPage. The latter can be modeled with the former (you can recalculate), but not the opposite as the latter assumes that each page has same number of items - this assumptions is no more with offset/limit approach.

alien-mcl:
This approach also removes the issue of page identifiers - skip/take will always be a number of collection elements.
The only thing I'm unhappy is the logic of the `item` - it can be either a single resource matching `manages` block, but in case it is not provided it could mean a single RDF statement.

tpluscode:
I don't see how either is more "generic" than the other. What's more, in my experience pageIndex/itemsPerPage approach is more common in API design.

I think we should not be deciding for and against either. Rather, provide a flexible enough set of primitives so that any design in possible to be described with Hydra.

Such is what I started discussing on #102, to base reduce the entire feature to IRI Templates. That way collection design is just one of possible applications of a more generic feature.

alien-mcl:
> I don't see how either is more "generic" than the other. 

As I wrote, you can express pageIndex/itemsPerPage with offset/limit, but not the opposite (limit can change on each request).

> What's more, in my experience pageIndex/itemsPerPage approach is more common in API design.

For desktop views, your right. But for mobiles, not really.
In streaming-line approach you'd favor skip/take as usually there is only `more` button, so the only thing to do us to add previous `take` and `skip` values and you've got another `skip` for next request (you would even omit `take` part and leave it for server to decide).

tpluscode:
> As I wrote, you can express pageIndex/itemsPerPage with offset/limit, but not the opposite (limit can change on each request).

That is not the point. If I have an API which has collection URLs in the form of `/collection{?pageIndex,pageSize}` there is nothing to express differently. Because whose responsibility it would be to translate one form to the other? The generic client's?

The design is what it is and we must allow to describe both and possible any other, even if it's going to be specific to the given API (if it should use custom descriptions).

> For desktop views, your right. But for mobiles, not really.

We should not discuss desktop vs mobile. We are designing a generic API Description vocabulary. If it favours one solution (skip/take) over another (page index/page size) then we have failed.

> In streaming-line approach you'd favor skip/take as usually there is only `more` button [...]

This argument is moot as such "streaming" it's already easier done by following the `next` link. This PR will not change it.

alien-mcl:
I've added another predicate for identifying a specific page. I've removed the range as the page identifier could be virtually anything (from number as in indexed pages to GUID).
I assumed that `hydra:take` will work as the page size. I was thinking about changing it to something more generic, i.e. `hydra:limit` - works better for both approaches.

angelo-v:
> I've added another predicate for identifying a specific page.

`hydra:page` still sounds, as if it links to some kind of page resource.

How about `hydra:pageReference`?

Or at least clarify the comment: "A server-known reference to a specific page to provide."

But again: Since it is a server-known reference, how can the paging be client-initiated? This actually only makes sense, if the client can assume that it is a page number it can simply increment. UUIDs etc. won't work out, because the client cannot know what to put in there.

tpluscode:
> This actually only makes sense, if the client can assume that it is a page number it can simply increment. UUIDs etc. won't work out, because the client cannot know what to put in there.

Hence my comments above and under #102. It is more important to ensure that the server can instruct the client what are the allowed value for the `pageReference`. In natural language those could be 

* natural number where `1 <= N < 100`
* a letter `[A-Za-z]` (think of a lexical index)
* anything else (requiring custom client code, which is fine)

To get this right we should also first ensure that we have a solid and share understanding for how the templates are supposed to be used. Please compare to #152

angelo-v:
> To get this right we should also first ensure that we have a solid and share understanding for how the templates are supposed to be used.

Since this is more complicated, I propose to only merge the skip/take approach for now (which adds at least some value for APIs using that approach) and work further on alternatives afterwards.

alien-mcl:
> Since it is a server-known reference, how can the paging be client-initiated?

Good point - skip/take approach is free of that issue as it relies only on number of items within the set (which will be always a natural number).

> anything else (requiring custom client code, which is fine)

This is related to #82 . We could allow to provide an explicit list of allowed literals or a link to the endpoint that would provide those.

tpluscode:
> Good point - skip/take approach is free of that issue as it relies only on number of items within the set (which will be always a natural number).

This is irrelevant. Like I said, not every API will want to follow the skip/take design and I'm pretty sure that it's a less common approach in practice outside of implementations close to a query language, like OData. I find it harmful that Hydra should inadvertently favour one possible approach over another.

This statement is also inaccurate. Just like with `pageReference` or any other alternative, it's not nearly enough to just tell the client

> these are the `skip` and `take` variables, deal with them

The server will most likely still impose restrictions, such as `Page size cannot be greater then 50`.

If we take a step back and look at ways to extend the vocab around templates, we will be able to represent any design

alien-mcl:
My proposal is:
- add `hydra:offset` (replacement for `hydra:skip` introduced here) with range of `xsd:nonNegativeInteger`
- add `hydra:limit` (replacement for `hydra:take` introduced here) with range of `xsd:nonNegativeInteger`
- add `hydra:pageIndex` with range of `xsd:nonNegativeInteger`
- add `hydra:pageReference` with range unspecified for future detailed specification (custom indices)

This way we'll cover 99% of paging use cases (pageIndex/limit fits not too bad, offset/limit fits well) and we're still enabled for future custom usages including GUID or letter indices. The latter will require some more concept i.e. variable constraints, allowed literals/individuals, etc.

spec/latest/core/core.jsonld
```diff
@@ -505,6 +509,36 @@
...
+    },
+    {
+      "@id": "hydra:pageIndex",
+      "@type": "rdf:Property",
+      "label": "page index",
+      "comment": "Instructs to provide a specific page of the collection at a given index.",
+      "vs:term_status": "testing"
+    },
+    {
+      "@id": "hydra:pageReference",
```
vddesai1871: Is it only me or `pageReference` still does not make much sense? Can't we just go with `offset`, `limit` and `pageIndex` for now and create another issue or PR before we add it to the vocab.

spec/latest/core/core.jsonld
```diff
@@ -505,6 +509,36 @@
...
+    {
+      "@id": "hydra:limit",
+      "@type": "rdf:Property",
+      "label": "take",
+      "comment": "Instructs to limit set only to N elements.",
+      "range": "xsd:nonNegativeInteger",
+      "vs:term_status": "testing"
+    },
+    {
+      "@id": "hydra:pageIndex",
```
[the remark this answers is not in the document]
alien-mcl: Ahh, right. Here it is

spec/latest/core/core.jsonld
alien-mcl: We wouldn't like to limit possible usages only to numeric pages - that's why `pageIndex`. I've recently added a `subPropertyOf` relation between `pageIndex` and `pageReference` - it feels natural this way.
Both predicates has similar semantics - here client you can point to a specific page of the set. The extra thing that comes with `pageIndex` is it's cleanness, which covers most use cases.
While it is not possible to provide i.e. min/max range or other means of expressing allowed values for `pageReference`, this predicate still can be used as it is now with some success.

alien-mcl:
Is it OK to merge this one? Are there any objection against this solution?

angelo-v:
> Is it OK to merge this one? Are there any objection against this solution?

I am fine with it.

tpluscode:
Yup, do it.

angelo-v:
@alien-mcl There is a problem with the rendering of html page:

hydra:pageIndex
Instructs to provide a specific page of the collection at a given index.

Range: undefined

**Subproperty of: [object Object]**

Status: testing

http://www.hydra-cg.com/spec/latest/core/#hydra:pageIndex

tpluscode:
@alien-mcl please turn @angelo-v's comment into a new issue

Issue #109, opened 2017-01-03
Use W3C HTTP RDF vocabulary


asbjornu:
Since there's a [HTTP vocabulary for RDF](https://www.w3.org/TR/HTTP-in-RDF/), wouldn't it be wise for Hydra to use it rather than define its own `hydra:method` and such? I know it's just a working draft, but perhaps Hydra's reliance on it may help nudge it in the right direction?

tpluscode:
Yup, I've just recently mentioned http vocab under @elf-pavlik's comment in #2.

elf-pavlik:
I think hydra JSON `@context` could simply map
```json
{
  "@context": {
    "http": "http://www.w3.org/2011/http#",
    "method": "http:methodName"
  }
}
```
Whoever works directly with Turtle would will know how to use prefixes and terms from different namespaces.

@asbjornu which other terms from *HTTP vocabulary for RDF* do you find useful for Hydra?

asbjornu:
@elf-pavlik, I haven't had a thorough look at it yet, so I'm not sure. But for every term that exists in the HTTP vocabulary, we should rather use that in Hydra than define our own. Whether we use the terms directly referencing the HTTP vocabulary or whether we remap them as you suggest is not that important to me. Either would work fine, I think.

lanthaler:
We discussed this before.. see e.g. Ruben's mail here: https://lists.w3.org/Archives/Public/public-hydra/2014Jun/0022.html

In my opinion reusing a vocabulary only makes sense if it being (widely) used elsewhere. I haven't seen any usage of the HTTP vocabulary so far and it isn't being maintained anymore. I would thus strongly prefer to keep the control of such a crucial aspect of Hydra within our group.

Btw. you certainly don't want to map the vocab to the `http` prefix :-)

elf-pavlik:
At some point I looked at *HTTP vocabulary for RDF* trying to find a way to describe HTTP redirects and manage them with REST API. I didn't find clear way to do that but I may pick up on this possibility at some point. AFAIK servers usually don't provide API to set HTTP Headers and satus code which resource will respond with.

tpluscode:
> AFAIK servers usually don't provide API to set HTTP Headers and status code

Not that they couldn't. It is something tha the HTTP API could be used for. But it seems very strict in its semantics.

lanthaler:
[...]

PR #185, opened 2019-02-11
Add support for describing headers

alien-mcl:
This is a derivative of a PR #182 that covers #99 :
- added terms for providing returned/expected headers
- added support for header value templating


spec/latest/core/core.jsonld
```diff
@@ -428,7 +434,7 @@
     {
       "@id": "hydra:IriTemplate",
       "@type": "hydra:Class",
-      "subClassOf": "hydra:Resource",
+      "subClassOf": ["hydra:Template", "hydra:Resource"],
```
angelo-v: Do we need to mention hydra:Resource here? A hydra:Template is a hydra:Class is a hydra:Resource.

spec/latest/core/core.jsonld
```diff
@@ -505,6 +511,55 @@
...
+      "rangeIncludes": ["hydra:Header", "hydra:HeaderTemplate"],
+      "vs:term_status": "testing"
+    },
+    {
+      "@id": "hydra:Template",
+      "@type": "rdfs:Class",
+      "label": "Template",
+      "comment": "Some abstract template.",
+      "vs:term_status": "testing"
+    },
```
tpluscode: I'd like to challenge the very need for an `abstract` template. It sounds like smuggling OOP terminology into the vocabulary which strikes me as awkward

spec/latest/core/core.jsonld
tpluscode: In my opinion the weirder part is actually introducing inheritance here. `IriTemplate` is not just any string template. It is meant as implementing the [rfc6570](https://tools.ietf.org/html/rfc6570). The changes here make the relation with its original RFC... complicated.

spec/latest/core/core.jsonld
```diff
@@ -485,7 +491,7 @@
       "@type": "rdf:Property",
       "label": "mapping",
       "comment": "A variable-to-property mapping of the IRI template.",
-      "domain": "hydra:IriTemplate",
+      "domain": "hydra:Template",
```
tpluscode: Regardless of the other template changes, I think this should remain an IRI Template. 

Or to test the logic, does it make sense to have a `HeaderTemplate` used with `rdf:Property`?

spec/latest/core/core.jsonld
```diff
@@ -505,6 +511,55 @@
...
+    },
+    {
+      "@id": "hydra:Header",
+      "@type": "rdfs:DataType",
+      "label": "Header",
+      "comment": "Specification of the header, including name and value(s).",
+      "vs:term_status": "testing"
+    },
+    {
+      "@id": "hydra:HeaderTemplate",
```
tpluscode: Still on the subject of headers, I think we're reinventing the wheel here.

What do you think about adopting the structure and terminology from actual [HTTP spec](https://tools.ietf.org/html/rfc7230#section-3.2) to describe the headers?

spec/latest/core/core.jsonld
angelo-v: rfc6570 is just the default:

> The literal's datatype indicates the template syntax; if not specified, hydra:Rfc6570Template is assumed.

And this quote only describes the [`hydra:template`](https://www.hydra-cg.com/spec/latest/core/#hydra:template) property, not the `hydra:IriTemplate`

Nevertheless I also have a bad feeling with this type of inheritance.

spec/latest/core/core.jsonld
angelo-v: This has been discussed already: https://github.com/HydraCG/Specifications/issues/109

> It seems that using unmaintained vocabulary that would in addition tightly couple hydra to a specific protocol is not a good idea. Hydra is meant to be protocol-agnostic, thus I don't see any option of using such vocabularies in the spec. -- https://github.com/HydraCG/Specifications/issues/109#issuecomment-452431098

Regarding this, should we support the description of (HTTP) Headers anyway?

tpluscode: Sorry, I did not mean to propose the HTTP vocabulary. Just to follow the actual protocol specification to describe headers

angelo-v: Sorry, misinterpreted that, should have followed the link ^^

spec/latest/core/core.jsonld
alien-mcl: I did not typed `hydra:Template` as any of the `hydra` types, thus to maintain backward compatibility, we have to point to `hydra:Resource` directly.

spec/latest/core/core.jsonld
alien-mcl: I'm opened for alternatives.

This approach is to have backward compatibility and ability to use `hydra:template` on both `hydra:IriTemplate` and newly introduced `hydra:HeaderTemplate`.

Another approach would be to introduce completely new `hydra:template` alternative for `hydra:HeaderTemplate`. I think it would be prone to errors.

Yet another approach would be to completely remove header templates at all. I believe templates could cover more complex header value scenarios, i.e. those for `PREFER`, where you provide some complex expressions that could be bound to variables.

spec/latest/core/core.jsonld
alien-mcl: Yep, RFC 6570 is not directly bound to IriTemplate. In RDF representation of hydra spec it is connected only via `seeAlso`, thus no strict data type is enforced.

spec/latest/core/core.jsonld
alien-mcl: Well, the `hydra:HeaderTemplate` would also need a `hydra:mapping`. I've just noticed that the range should be also decoupled from IriTemplateMapping.

The thing is, IriTemplate implies ... an IRI, which is not the case here. For headers (or body in future??) that would be completely different templating mechanism (i.e. razor :P)

spec/latest/core/core.jsonld
alien-mcl: The idea with template in headers was to allow server to provide i.e. complete request template, including headers and/or body so the client just binds that template with variable values and sends back.

angelo-v:
LGTM

PR #195, opened 2019-05-15
add section about manages block

### LINKED ISSUE #41: Discuss how usage of Collections "breaks" relationships
On Mar 18, 2014, at 3:13 PM, [Markus Lanthaler wrote](http://lists.w3.org/Archives/Public/public-hydra/2014Mar/0067.html):

> On Saturday, March 15, 2014 1:03 AM, Gregg Kellogg wrote:
> 
> > For that matter, if I have a property that I've described as being a
> > hydra:Link, the value of that property is presumably to a resource
> > returning a collection, rather than the resource directly described as
> > being, e.g., a colleague. If you build a graph from following these
> > indirections, it leads to a graph which co-mingles entities of type
> > hydra:Collection and schema:Person, in this example. I've also
> > addressed this by using different properties (schema:colleagues vs
> > schema:colleague), but this is probably controversial.
> 
> Yeah, you know my opinion about that :-) IMO, the consumer needs to be able
> to reliably work with both forms (and cases that the plural instead of the
> singular form and vice versa). If so, it actually becomes simpler to just
> have to deal with a single property instead of two.
> 
> > Here's an example:
> > 
> > ```
> >  </GreggKellogg> a schema:Person;
> >    schema:colleagues </GreggKellogg/colleagues>;
> >    schema:colleague </MarkusLanthaler>, ...
> >  </GreggKellogg/colleagues> a hydra:Collection;
> >    hydra:member </MarkusLanthaler>, ...
> >  </MarkusLanthaler> a schema:Person .
> > ```
> 
> Yeah.. in principle this is quite similar as lists in RDF. They also
> introduce intermediary nodes which "break" the direct link between the
> nodes. It might make sense to discuss this in the spec.

[Gregg also suggests](http://lists.w3.org/Archives/Public/public-hydra/2014Mar/0072.html) to also include "some best practices for dealing with the potential confusion"

### LINKED ISSUE #60: Find a better name for the "manages block"
Issue #41 was resolved by introducing a new property `hydra:manages` that can be used to (partially) describe the collection's members. The range of `hydra:manages` has been refered to as "manages block" in discussions till now but for the specification a better, more descriptive term is needed. Proposals are:
- TripleTemplate 
- Relationship 
- RelationshipTemplate 

It also has to be decided if it is worth to give that term a URL (will clients benefit?) or whether it is purely intended to be used in prose.

### LINKED ISSUE #126: Simplify discovery of collections that contain resources of a certain type

### LINKED ISSUE #150: Rename manages to memberAssertion
Can we please rename the `manages` property wrt. collections to `contains`? I just find the latter to be so much more self-descriptive than the former.

### LINKED ISSUE #194: Manages block usage for subject vs object
## Description

I've been working on implementing "manages block" in Alcaeus and I conclude that it can have two possible usages as currently defined (well, implicitly).

First is to assert member types, as in the #126 by @elf-pavlik:

```json
{
  "manages": {
    "rdf:predicate": "rdf:type",
    "rdf:object": "example:Class"
  }
}
```

The other is to fix the "broken relationships" as discussed in #41 among other threads

```json
{
  "manages": {
    "rdf:predicate": "foaf:knows",
    "rdf:subject": { "@id": "GreggKellogg" }
  },
  "member": [
     { "@id": "Markus" }
  ]
}
```

## Questions

Now, does it actually make sense to include all three `subject`, `predicate` and `object` in a single manages block?

tpluscode:
## Summary

I added the "manages" to the spec, finally 🎉 

## More details

We can close #41 which is directly related to this proposal and has already been "resolved".
For same reason we should close #126.

Fixes #194 as it aims to clarify the semantics and usage

I would like to close #150 and close #60 - the name "manages block" may should a little alien but I feel like it's a Hydra-specific enough construct to let it be justified. It's been around long enough too that it settled IMO.

## Related cookbook PR 

Previously partially addressed as #132

### DIFF
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -854,7 +854,7 @@
 
     <p>In the above example, adding a <code>manages</code> node to the collection instructs the
       client that every member of this collection is linked to the <code>subject</code>
-      by the <code>property</code>. It could be written a SPARQL triple pattern below, where
+      by the <code>property</code>. It could be written as a SPARQL triple pattern below, where
       <code>?m</code> would be substituted by each member of the collection.</p>
 
     <pre class="example nohighlight" data-transform="updateExample">
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -868,7 +868,7 @@
       each expressing different relations between the collection members and other resources.</p>
 
     <p class="note">It's important to point out that the <code>subject</code>, <code>property</code>
-      and <code>object</code> predicate are defined within the Hydra namespace and are not
+      and <code>object</code> predicates are defined within the Hydra namespace and are not
       <a href="https://www.w3.org/TR/rdf-schema/"><code>rdf</code></a> terms.</p>
   </section>
 
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -748,7 +748,7 @@
     described somewhere in the beginning?</p>
 
   <section>
-    <h3>Collections</h3>
+    <h3 name="collections">Collections</h3>
 
     <p>In many situations, it makes sense to expose resources that reference
       a set of somehow related resources. Results of a search query or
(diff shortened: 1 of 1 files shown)


angelo-v:
Thanks for clarifying that in the spec. It took me quite a while to figure it out myself when starting with hydra!

tpluscode:
Haha, took me even longer 😬

spec/latest/core/index.html
```diff
@@ -823,6 +823,55 @@ <h3>Collections</h3>
...
+            "@id": "/comments/429"
+          }
+        ]
+      }
+      -->
+    </pre>
+
+    <p>In the above example, adding a <code>manages</code> node to the collection instructs the
+      client that every member of this collection is linked to the <code>subject</code>
+      by the <code>property</code>. It could be written a SPARQL triple pattern below, where
```
vddesai1871: A minor typo *"It could be written as a SPARQL triple pattern"*
btw thanks for this, earlier I went through all PRs related to *manages* block but couldn't understand much, now it makes perfect sense.

spec/latest/core/index.html
```diff
@@ -823,6 +823,55 @@ <h3>Collections</h3>
...
+      <http://api.example.com/an-issue> <http://api.example.com/vocab#comment> ?m
+      -->
+    </pre>
+
+    <p>A manages block MUST use two and only two of the <code>subject</code>, <code>property</code>
+      and <code>object</code> predicates. There MAY be more than one such manages blocks,
+      each expressing different relations between the collection members and other resources.</p>
+
+    <p class="note">It's important to point out that the <code>subject</code>, <code>property</code>
+      and <code>object</code> predicate are defined within the Hydra namespace and are not
```
alien-mcl: I think that `predicate` should be replaced with `predicates`

tpluscode:
Thank you @alien-mcl @vddesai1871, I applied your suggestions

asbjornu:
I’m still not content with the word `manages`. A collection does not “manage” anything, since it’s not a “manager”. A collection “contains”, “holds” “consists of”, “comprises of” or “encloses” items.

tpluscode:
> A collection “contains”, “holds”, “consists of”, “comprises of” or “encloses” items.

The manages block is about neither of the relation the you give as example. Instead it's a way to indirectly make statements about the members. 

If you have some suggestion I propose that you reopen #60. That issue is fairly pristine so we could revisit this decision.

alien-mcl:
> A collection “contains”, “holds”, “consists of”, “comprises of” or “encloses” items.

But a resource you access that collection through actually does manage that collection. We're not talking about a collection, but it's physical representation through an API. 

But I agree with @tpluscode - feel free to reopen #60

asbjornu:
I'll reopen #150 since I created it. I can't open issues I haven't created.

asbjornu:
Hm, no. I can't reopen #150 either. The restrictions on this repository is driving me mad.

PR #198, opened 2019-06-18
TemplatedLink cannot formally have supportedOperation

alien-mcl:
## Summary

The `hydra:TemplatedLink` is not formally considered a `hydra:Link` so it cannot formally have any of the properties that are _allowed_ for `hydra:Link`. While in RDF each property can be used on every resource, it is still preferred to explicitly define some of the relations between terms so the client implementation is easier and the data model is stable.

## More details

The `hydra:TemplatedLink` is sub-classed from `hydra:Link` in this template making it formally also a `hydra:Link` which enables it to all properties define for `hydra:Link` with either `rdfs:domain` or `schema:domainIncludes`.

### DIFF
--- a/spec/latest/core/core.jsonld
+++ b/spec/latest/core/core.jsonld
@@ -450,7 +450,7 @@
     {
       "@id": "hydra:TemplatedLink",
       "@type": "hydra:Class",
-      "subClassOf": [ "hydra:Resource", "rdf:Property" ],
+      "subClassOf": [ "hydra:Resource", "hydra:Link", "rdf:Property" ],
       "label": "Templated Link",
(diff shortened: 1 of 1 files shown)


alien-mcl:
I've replaced approach in this PR with last commit. This is due to potentially issue raised by @angelo-v on one of the occasions. Previous approach sub-classed `hydra:TemplatedLink` from `hydra:Link`. While convenient, it could cause issues with interpretations of the resource on the right side of the relation.

It could be considered a link, while it should - it will be an IRI template that needs to be resolved first (becoming a real link).

New approach feels safer from that point of view still providing required semantics.

Issue #199, opened 2019-07-17
Another take on non-RDF payloads (aka file upload)


tpluscode:
## Describe the requirement

We've approached the problem of non-RDF media types a few times already. Unfortunately it seems that each time it was not focused enough. Either mixed with collections (#187) or lacking broader context (#186).

Looking back at both, I think they are on track, but need a little more refinement.

_For this issues, I'd like to focus on `expects` only and not returned representations_

## Hydra-agnostic example

I would distinguish 3 kinds of requests coming to a Hydra API:

1. RDF payloads - such that are currently described by `expects` and `Class`.

2. Non RDF-payloads

   Directly uploading an image instead of RDF:

   ```http
   POST /movie/123/poster-image HTTP/2
   Content-Type: image/png
   
   ...File bytes...
   ```

3. `multipart/form-data`

   Submitting multiple images **and** RDF data:

   ```http
   PUT /movie/123/image-gallery HTTP/2
   Content-Type: multipart/form-data; boundary=----hydra-content
   
   ----hydra-content
   Content-Disposition: form-data; filename="poster.png"
   Content-Type: image/png
   
   ...Poster image...
   ----hydra-content
   Content-Disposition: form-data; filename="cast.jpeg"
   Content-Type: image/jpeg
   
   ...Image of actors...
   ----hydra-content
   Content-Type: application/ld+json
   
   {
     "@type": "mov:Gallery",
     "description": {
       "@value": "Pictures for movie /movie/123"
     }
   }
   ----hydra-content
   ```

Hydra should allow describing operations which `expect` both kinds of file uploads. 

## Proposed solutions

It is important to keep support for the current `expect` semantics. 

I propose that we extend the existing structure with a media-type description. Unfortunately it is not possible to have it both ways without revolutionising the structure, so the vocab will have to remove `rdfs:range` from `expects` and use `schema:rangeIncludes` instead.

```diff
{
  "@id": "hydra:expects",
-  "range": "hydra:Class",
+  "schema:rangeIncludes": [
+    "hydra:Class",
+    "hydra:RequestSpecification"
+  ]
},
```

### Example of `hydra:Class` usage

```json
{
  "@type": "Operation",
  "expects": {
    "@type": "RequestSpecification",
    "content": {
      "@type": "SupportedClassContent",
      "class": "mov:Movie"
    }
  }
}
```

This would be equivalent to `"expects": "mov:Movie"` and both should be supported at least for a while.

### Example of non-RDF payload

```json
{
[...]

PR #200, opened 2019-09-09
Enable hydra to expect and return non RDF resources

alien-mcl:
## Summary

I've changed the `rdfs:range` of `hydra:expects` and `hydra:returns` predicates from `hydra:Class` to `hydra:Resource` and introduced `schema:rangeIncludes` for both.
I've also mentioned that fact in the spec.

## More details

This PR addresses issues #22 and #199 but does *NOT* resolves them. It is also an alternative approach to PRs #186 and #187.

Sole purpose is to enable the vocabulary to expect and return non-RDF resources. Previous specification placed a `hydra:Class` in the range the proper predicates making it difficult to use other types of resources.
Unfortunately, modification within the `rdfs:range` of the aforementioned predicates may introduce possible issues within current implementations and specifications built on top of the vocabulary. Still, it seems to be the least intrusive method due to facts:
- the `hydra:Class` used previously is still compliant with newly introduced `hydra:Resource`
- `schema:rangeIncludes` introduced does not apply strict RDF type entailing rules yet still provides semantic hints on possible values.

### DIFF
--- a/spec/latest/core/core.jsonld
+++ b/spec/latest/core/core.jsonld
@@ -253,7 +253,8 @@
       "label": "expects",
       "comment": "The information expected by the Web API.",
       "domain": "hydra:Operation",
-      "range": "hydra:Class",
+      "range": "hydra:Resource",
+      "rangeIncludes": ["hydra:Resource", "hydra:Class"],
       "vs:term_status": "testing"
     },
     {
@@ -262,7 +263,8 @@
       "label": "returns",
       "comment": "The information returned by the Web API on success",
       "domain": "hydra:Operation",
-      "range": "hydra:Class",
+      "range": "hydra:Resource",
+      "rangeIncludes": ["hydra:Resource", "hydra:Class"],
       "vs:term_status": "testing"
     },
     {
--- a/spec/latest/core/core.jsonld
+++ b/spec/latest/core/core.jsonld
@@ -77,7 +77,8 @@
     "subClassOf": { "@id": "rdfs:subClassOf", "@type": "@vocab" },
     "subPropertyOf": { "@id": "rdfs:subPropertyOf", "@type": "@vocab" },
     "seeAlso": { "@id": "rdfs:seeAlso", "@type": "@id" },
-    "domainIncludes": { "@id": "schema:domainIncludes", "@type": "@id" }
+    "domainIncludes": { "@id": "schema:domainIncludes", "@type": "@id" },
+    "rangeIncludes": { "@id": "schema:rangeIncludes", "@type": "@id" }
   },
   "@id": "http://www.w3.org/ns/hydra/core",
   "@type": "owl:Ontology",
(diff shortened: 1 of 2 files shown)


spec/latest/core/core.jsonld
```diff
@@ -253,7 +254,8 @@
       "label": "expects",
       "comment": "The information expected by the Web API.",
       "domain": "hydra:Operation",
-      "range": "hydra:Class",
+      "range": "hydra:Resource",
+      "rangeIncludes": ["hydra:Resource", "hydra:Class"],
```
tpluscode: I think it's redundant to repeat `hydra:Resource` here. In fact, a `Resource` itself is not useful as an object of `expects`/`returns`. It has to be something more specific, like `Class`

```suggestion
      "rangeIncludes": ["hydra:Class"],
```

spec/latest/core/core.jsonld
```diff
@@ -262,7 +264,8 @@
       "label": "returns",
       "comment": "The information returned by the Web API on success",
       "domain": "hydra:Operation",
-      "range": "hydra:Class",
+      "range": "hydra:Resource",
+      "rangeIncludes": ["hydra:Resource", "hydra:Class"],
```
tpluscode: Ditto
```suggestion
      "rangeIncludes": ["hydra:Class"],
```

spec/latest/core/index.html
```diff
@@ -533,7 +533,11 @@ <h3>Documenting a Web API</h3>
       certain operations depend on the permissions of the current user. It
       makes it, e.g., possible to show a "delete" link only if the current
       user has the permission to delete the resource. Otherwise, the link
-      would simply be hidden in the representation.</p>
+      would simply be hidden in the representation.<br />
```
```suggestion
      would simply be hidden in the representation.</p>
```

spec/latest/core/index.html
```diff
@@ -533,7 +533,11 @@ <h3>Documenting a Web API</h3>
...
+      Example shown below describes the operation's expected and returned
```
```suggestion
    <p>Example shown below describes the operation's expected and returned
```

spec/latest/core/core.jsonld
alien-mcl: I don't agree. I think it's useful to have both predicates to contain `hydra:Resource`. Previously used `rdfs:range` is more for compatibility, but I'd like to use `schema:rangeIncludes` in future implementations - it has weaker semantics.

I'd leave it as I suggested, at least untill some more specific alternatives are available.

asbjornu: Isn't `hydra:Resource` what enables non-RDF payloads?

alien-mcl: @asbjornu, `schema:rangeIncludes` doesn't have that strong semantics as `rdfs:range`. Putting only `hydra:Class` inside `schema:rangeIncludes` doesn't infer no other types can be used - it's more of a hint (well the name has it - range includes, but not restricts). Still, I'd prefer to have both so clients doesn't have to concatenate both `rdfs:range` and `schema:rangeIncludes` to discover what's actually allowed.

asbjornu: I see. Thanks for the explanation, @alien-mcl!

tpluscode: But `Resource` is range includes is actually not truthful. Most classes of Hydra are subclasses of `hydra:Resource` but they will not make sense as `expects`/`returns`. This hint is misleading.

Issue #202, opened 2019-11-26
Clarify behaviour when ApiDocumentation resource contains more than just itself


tpluscode:
## Description

Having talked on today's conference call, @serialseb proposed to explicitly allow the `ApiDocumentation` to have more than just Hydra-related triples. In the provided example scenario and API may also embed additional RDFS vocabulary terms which the client will need to fully understand the data structures 

_(@serialseb, please do open a separate issue in which you provide information on how and why this information would be useful)_

## Example

Written in turtle, that would simply be multiple subgraphs.

```
# the API Doc itself with all the standard bits
@base <http://some.app/> .
@prefix api: <http://some.api/> .
@prefix ex: <http://ontology.example/> .
@prefix hydra: <http://www.w3.org/ns/hydra/core#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .

<api> 
  a hydra:ApiDocumentation ;
  hydra:supportedClass api:ClassOne, api:ClassTwo .

# Anything else
ex:SomeType a rdfs:Class .
```

[...]

PR #203, opened 2020-01-20
Clarify behaviour when ApiDocumentation resource contains more than just itself

alien-mcl:
## Summary

This pull request addresses an issue #202 Clarify behaviour when ApiDocumentation resource contains more than just itself.

## More details

There are some additional examples and paragraphs describing on how to tackle RDF graphs with additional tripples that are not diretly related to an API documentation.
A couple of editortial issues were also fixed in the process.

### DIFF
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -566,7 +566,7 @@
       names. Client SHOULD apply respective header semantics when creating
       or receiving a request natural for the protocol in use.</p>
 
-    <pre class="example nohighlight" data-transform="returnsHeader"
+    <pre class="example nohighlight" data-transform="updateExample"
          title="Making a use of a returned header">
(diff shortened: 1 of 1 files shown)


spec/latest/core/index.html
```diff
@@ -627,10 +627,64 @@ <h3>Documenting a Web API</h3>
 
     <p>Aforementioned <i>title</i> and <i>description</i> SHOULD take precedence
       over standard <i>rdfs:label</i> and <i>rdfs:comment</i>.</p>
+
+    <p>There is one more relish related to how Linked Data works. Consider the
```
asbjornu: Relish?

Relish

alien-mcl: Why not? :). Anyway, I've reworded it

Issue #196, opened 2019-05-21
Supported Operations on properties


angelo-v:
It is proplematic to define supported operations on a `hydra:property`, when re-using vocabularies, like on http://schema.org/events below:

```json
"supportedProperty": [
    {
      "property": {
        "@id": "schema:events",
        "supportedOperation": [
          {
            "@id": "_:some_operation"
          }
        ]
      }
    }
  ]
```
This now states that `schema:events` does support `_:some_operation`, which is not true in general but only specific to this API.

It would be better to move anything API specific to the level of `supportedProperty`:

```json
  "supportedProperty": [
    {
      "property": {
        "@id": "schema:events"
      },
      "supportedOperation": [
        {
          "@id": "_:some_operation"
        }
      ]
    }
  ]
```

This is e.g. already the case for `required`, `readable` and `writable`.

alien-mcl:
Hmm. I think there is some misunderstanding of the spec. Your second snipped is exactly as the vocab says. Take a look at the raw JSON-LD vocabulary and you'll see that mentioned predicates (`required`, `readable`, and `writable`) are pointing as their domains 'SupportedProperty'.

As for the 'supportedOperation' - it points to 'Class' as it's domain and while `SupportedOperation` is of type `Class` you can safely use `supportedOperation` with it.

This is exactly to avoid adding unnecessary statements to the property itself.

Correct syntax would be:
```json
"supportedProperty": [
    {
      "property": {
        "@id": "schema:events"
      },
      "required": true,
      "supportedOperation": [
        {
          "@id": "_:some_operation"
        }
      ]
    }
  ]
```

tpluscode:
Clearly, this is not obvious. Even the ApiDoc of the issue tracker demo does it "wrong": http://www.markus-lanthaler.com/hydra/event-api/vocab

We definitely have to make this clear in the spec

angelo-v:
So wich on would get the `@type: Link` then? the supported property, or the property?

alien-mcl:
>Clearly, this is not obvious.

Well, there is also an image that shows it on top of the spec, but it seems that indeed it should be stated more precisely.

alien-mcl:
It seems that my last `correct` example is not quite `correct`.
Indeed, `hydra:supportedOperation` is used in some of the examples directly on the property rather than the supported property and the spec does not touch that matter directly. Also the raw RDF representation of the vocabulary is not that clear about it.

[...]

PR #204, opened 2020-02-26
Allowed supportedOperation on SupportedProperty.

alien-mcl:
## Summary

Allowing _supportedOperation_ predicate on _SupportedProperty_ instances.
It also marks several terms as *stable*.

## More details

This PR fixes issue #196. In general, there were doubts on how to (if possible) attach a supported operation to the property. There were few examples that did this by using _supportedOperation_ on the property itself, but this was incorrect neither from the raw range/domain point of view defined within the RDF spec nor semantic understanding of such approach.

This PR enables _supportedOperation_ to be formally attachable to instances of _SupportedProperty_. This way this intermediate structure can hold any API specific details preventing from leaking it to externally defined properties.

### DIFF
--- a/spec/latest/core/core.jsonld
+++ b/spec/latest/core/core.jsonld
@@ -98,7 +98,7 @@
       "subClassOf": "rdfs:Resource",
       "label": "Hydra Resource",
       "comment": "The class of dereferenceable resources.",
-      "vs:term_status": "stable"
+      "vs:term_status": "testing"
     },
     {
       "@id": "hydra:Class",
@@ -106,7 +106,7 @@
       "subClassOf": [ "hydra:Resource", "rdfs:Class" ],
       "label": "Hydra Class",
       "comment": "The class of Hydra classes. Hydra classes and their instances are dereferenceable resources.",
-      "vs:term_status": "stable"
+      "vs:term_status": "testing"
     },
     {
       "@id": "hydra:Link",
@@ -114,7 +114,7 @@
       "subClassOf": [ "hydra:Resource", "rdf:Property" ],
       "label": "Link",
       "comment": "The class of properties representing links.",
-      "vs:term_status": "stable"
+      "vs:term_status": "testing"
     },
     {
       "@id": "hydra:apiDocumentation",
@@ -123,7 +123,7 @@
       "comment": "A link to the API documentation",
       "range": "hydra:ApiDocumentation",
       "domain": "hydra:Resource",
-      "vs:term_status": "stable"
+      "vs:term_status": "testing"
     },
     {
       "@id": "hydra:ApiDocumentation",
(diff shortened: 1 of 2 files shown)


spec/latest/core/core.jsonld
```diff
@@ -98,23 +98,23 @@
       "subClassOf": "rdfs:Resource",
       "label": "Hydra Resource",
       "comment": "The class of dereferenceable resources.",
-      "vs:term_status": "testing"
+      "vs:term_status": "stable"
```
tpluscode: All of those changes appear unrelated to the matter at hand IMO

spec/latest/core/index.html
```diff
@@ -601,6 +601,14 @@ <h3>Documenting a Web API</h3>
       for the operation invocation.
     </pre>
 
+    <p>To wrap up everything altogether, it is also possible to attach atomic
+      operations supported by, well, supported property itself. This might
+      come in handy for scenarios, when resource can be partially modified.
```
tpluscode: This is a little unclear without an example TBH. 

I used this feature for example on a "typed" collection link.

```turtle
# in data
<> a api:Entrypoint ; api:dataSheetCollection <data-sheets> .

# in API Documentation
api:Entrypoint
  hydra:supportedProperty [
    hydra:property api:dataSheetCollection ;
  ] .

api:dataSheetCollection
  a                        hydra:Link ;
  rdfs:range               <api#DataSheetCollection> ;
  hydra:supportedOperation _:PostToCollection .
```

This way my client dereferences `<>` (entrypoint) and discovers that it can POST to `<data-sheets>`.

----

So now that would change like so: 

```diff
api:Entrypoint
  hydra:supportedProperty [
    hydra:property api:dataSheetCollection ;
+   hydra:supportedOperation _:PostToCollection ;
  ] .

-api:dataSheetCollection
-  a                        hydra:Link ;
-  rdfs:range               <api#DataSheetCollection> ;
-  hydra:supportedOperation _:PostToCollection .
```

spec/latest/core/core.jsonld
alien-mcl: Ok. I didn't want to create a separate PR that is unrelated to any issue, but let it be that way. I'll update it ASAP.

spec/latest/core/index.html
alien-mcl: Good point. I'll make required amendments to over that scenario.
In general, sole purpose was to make examples like `<Me> schema:knows <You>` obsolete as `schema:knows` is not `hydra:Link`. But your example clearly states that `api:dataSheetCollection` is a `hydra:Link`, thus it can have a `hydra:supportedProperty` attached.

alien-mcl: Done. I was wondering whether providing both examples could be useful.

PR #206, opened 2020-03-11
Add QPF spec

rubensworks:
This PR adds a spec document for Quad Pattern Fragments, an extension to the Triple Pattern Fragments spec.

Since TPF spec already resides in this repo, QPF also fits in nicely here.

\cc @RubenVerborgh

### DIFF
--- a/spec/latest/quad-pattern-fragments/index.html
+++ b/spec/latest/quad-pattern-fragments/index.html
@@ -1,0 +1,442 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>Quad Pattern Fragments</title>
+  <meta charset="utf-8">
+  <script type="text/javascript">
+//<![CDATA[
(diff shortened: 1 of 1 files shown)


tpluscode:
I stand by what I said earlier, that IMO neither TPF nor QPF fits in this repo really well

RubenVerborgh:
@tpluscode Happy to move it out, on the condition you could set up a permanent redirect at the URLs, such that we don't break links. Would that be possible?

alien-mcl:
I personally has no issues with having either TPF nor QPF in our repo. I believe TPF was the first to use hydra under the hood. I'd just love to see you guys a little bit more active, especially in terms of discussing and resolving all the issues raised.

tpluscode:
Ok, I rest my case.

RubenVerborgh:
Thanks for the input. Could you decide this with the Hydra group and formulate an official answer, so we know what next steps to take?

rubensworks:
> Could you decide this with the Hydra group and formulate an official answer, so we know what next steps to take?

Any updates on this?
Let me know if any more input is needed from our end.

alien-mcl:
As I wrote earlier - I see no need to separate specs and I didn't raise that topic neither on mailing list nor GH. Unless it is your intent, I do not see need to involve whole community (of which you're also part of) and formulate official answers.
I did some preparations via _Projects_ view to organize stuff related to each spec, so we can see which issue is relevant in which area. I have only wish to see more activity from you guys. I just remember, that some time ago I tried to close some of TPF related issues as abandoned and I was asked not to do so. Unfortunately, without any further discussions taken to reach any kind of resolution.

RubenVerborgh:
Thanks all.

We have decided that all LDF specifications will be hosted on linkeddatafragments.org in the future. We will make PRs to set up the redirects for the existing specs when the migration is complete.

Issue #207, opened 2020-03-17
Discovering IriTemplate and others


tpluscode:
## Description

I have been discussing with @bergos recently the usage of `IriTemplate` and `TemplatedLink`.

### How I thought about them

I have always only considered the template to be directly attached to the link in question. Take a link to lookup (or create) a resource by appending a slug. Let it be a Person

```turtle
</people>
  a hydra:Collection ;
  api:personByName [
    a hydra:IriTemplate ;
    hydra:template "/people/{name}"
  ]
```

Where obviously `api:personByName a hydra:TemplatedLink` .

### How @bergos thinks about them

It never occurred to me however that the occurrence of `hydra:IriTemplate` could also be placed in the `ApiDocumentation` graph. 

```turtle
</people>
  a hydra:Collection ;
  api:personByName api:PersonByNameTemplate .
```

And in the api documentation resource a client would find `api:PersonByNameTemplate` defined as above.

---

While not technically forbidden, the spec is not explicit here either, is it? I think it totally makes to make it clear that any Hydra term which is valid in the resource, might also be only referenced and the client should look for it in the Api Documentation to find more about it.

bergos:
Thanks for creating the issues!

Some thoughts about the conflict I always felt when I had to put the template either into the resource data or API documentation and why #207 together with #208 could fix it:

Besides the IriTemplates, everything in Hydra can be separated very well in resource data and API documentation. IriTemplates don't fit into only one of the two categories. The template contains the IRI of the resource, so it's somehow connected to the resource. But the benefit of the template data comes from using the API. Therefore the question is can it be shifted in one or the other direction? Using the template without using the API will not work. Shifting it towards the API documentation direction could work if it would be possible to remove the resource part of the IRI. Reducing the template to a relative part and resolve the full IRI with the resource IRI, could fix that problem. Then we would have a clear separation: The resource IRI comes from the resource data and the variable part comes from the API documentation.

alien-mcl:
Ok - this looks like a thing that could be put in the specification as a separate paragraph.

I think we may end up with something defining how a client should behave in these circumstances:
- in case of an a right hand side related resource is not defined and it is supposed to be an IriTemplate, client SHOULD look in the API documentation for definition
- in case it is still undefined and the host name of the URL (not IRI!) of this resource matches the host name of the obtained document, client SHOULD dereference that URL and look for for definition
[...]

PR #209, opened 2020-03-29
Issue 207 Discovering IriTemplate and others

alien-mcl:
I've added a paragraph describing on how missing resources should be discovered, which should resolve issue #207.

In general, the paragraph describes a simple algorithm of how a generic client should behave in specific circumstances related to _IriTemplate_ scenarios. There are some examples of payloads and resulting outcome of client actions.

### DIFF
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -1078,6 +1078,15 @@
       }
       -->
     </pre>
+
+    <p>where</p>
+
+    <ul>
+      <li>resource <i>http://api.example.com/people</i> should have an <i>IriTemplate</i> available
+        as there is a complete definition of the template available at <i>http://api.example.com/doc/</i>.</li>
+      <li>resource <i>http://api.example.com/events</i> should not have an Iri template exposed as there
+        are no additional details available, neither in the initial resources' payload nor in the API documentation.</li>
+    </ul>
   </section>
 
   <section>
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -1024,7 +1024,7 @@
       an <i>IriTemplate</i> expected as a related resource, client may
       discover no additional statements describing it. These rules should
       be considered when working with <i>IriTemplate</i>s (and other
-      resources resources in general):
+      hypermedia resources in general):
     </p>
 
     <ul>
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -1017,6 +1017,67 @@
       }
       -->
     </pre>
+
+    <p>Due to fact that hydra is on top of an RDF, which is effectively a graph,
+      it may happen that a related resource (the one on the right-hand side of the
+      relation) may not be fully described in the resource's payload. In case of
+      an <i>IriTemplate</i> a client could find out a related resource should be
+      a resource carrying a <i>template</i> predicate, but no additional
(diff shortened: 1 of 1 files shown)


spec/latest/core/index.html
```diff
@@ -1017,6 +1017,76 @@ <h3>Templated Links</h3>
       }
       -->
     </pre>
+
+    <p>Due to fact that hydra is on top of an RDF, which is effectively a graph,
```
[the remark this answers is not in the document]
tpluscode: Let me "rephrase"
```suggestion
    <p>Due to fact that hydra is on top of an RDF, which is a graph,
```

spec/latest/core/index.html
```diff
@@ -1018,28 +1018,29 @@ <h3>Templated Links</h3>
...
-      a resource carrying a <i>template</i> predicate, but no additional
-      description is available. These rules should be considered when working with
-      <i>IriTemplate</i>s (and other hypermedia resources in general):
+    <p>Due to fact that hydra is on top of an RDF, which is a graph,
+      it may happen that a related resource (an object of the relation)
+      may not be fully described in the resource's payload. In case of
+      an <i>IriTemplate</i> expected as a related resource, client may
+      discover no additional statements describing it. These rules should
+      be considered when working with <i>IriTemplate</i>s (and other
+      hypermedia resources in general):
```
tpluscode, line 1027:
```suggestion
      be considered when working with <i>IriTemplate</i>s (and other
     resources in general):
```

How about making this apply not only to hypermedia resource but any resource?

spec/latest/core/index.html
```diff
@@ -1018,28 +1018,29 @@ <h3>Templated Links</h3>
...
     </p>
 
     <ul>
-      <li>in case of a right-hand side related hypermedia resource does not have all
+      <li>in case of an object expected to be a hypermedia resource does not have all
         the necessary statements for which it is a subject, client SHOULD look in the
-        API documentation for more details.</li>
-      <li>in case the mentioned resource after consulting an API documentation still
+        API documentation for more details</li>
+      <li>in case the mentioned object after consulting an API documentation still
```
tpluscode, line 1034:
```suggestion
      <li>in case the mentioned object, after consulting an API documentation, still
```

spec/latest/core/index.html
```diff
@@ -1017,6 +1017,77 @@ <h3>Templated Links</h3>
...
+
+    <ul>
+      <li>in case of an object expected to be a hypermedia resource does not have all
+        the necessary statements for which it is a subject, client SHOULD look in the
+        API documentation for more details</li>
+      <li>in case the mentioned object after consulting an API documentation still
+        does not have all the necessary statements for which it is a subject and both
+        mentioned object's Url and Url of the initially obtained resource has the
+        same scheme and authority (by means of RFC 3986 sections 3.1 and 3.2), client
+        SHOULD dereference that Url</li>
```
tpluscode, line 1038:
```suggestion
        same scheme and authority (by means of RFC 3986 sections 3.1 and 3.2), client
        SHOULD dereference that Url. If the resource does not have the same scheme and authority,
        the client MAY choose to dereference it (for example if the resource originates from another
        API well-known to the client)</li>
```

I propose to add a `MAY` option for the client to dereference resources from other APIs if they so choose

Issue #208, opened 2020-03-17
Relative Template URIs


tpluscode:
## Description

I would also clarify the intended behavior around relative IRI Templates.

Especially in the context of #207 it might be a neat trick to be able to reuse same template for multiple links.

Consider two collection resources:

```http
GET /people/

<http://example.com/people/>
  a hydra:Collection ;
  api:newPerson api:SubResourceTemplate .
```

```http
GET /events/

<http://example.com/events/>
  a hydra:Collection ;
  api:newEvent api:SubResourceTemplate .
```

And an API Documentation

```turtle
api:SubResourceTemplate
  a hydra:IriTemplate ;
  hydra:template "{name}" ;
  hydra:mapping [
    a hydra: IriTemplateMapping ;
    hydra:variable "name" ;
    hydra:property schema:name ;
    hydra:required true
  ] .
```

## What's missing

Notice that the template is reused for two completely unrelated properties. To make this work the client would construct the template and apply it to a base URI. I expect that URI to be the parent's identifier.

In the example above that would effectively mean `http://example.com/people/{name}` and `http://example.com/events/{name}` respectively

Does this make sense?

## Caveat

The above will not work as expected if the collection is identified like `http://example.com/people` (no trailing slash). Will result in 

```diff
-http://example.com/people/foo
+http://example.com/foo
```

which might be unexpected...

alien-mcl:
Ok, I gave it some more thoughs. In general, everything looks promising, but as usual, I've tried to crash-test the idea and here are some of my findings.

Consider example:
```http
GET /api/people?search=term

</api/people> a hydra:Collection ;
  api:newPerson api:SubResourceTemplate .
```

What is a _parent_ here?

Another example, a bit more ephemeral:
```http
GET /api/people/john/knows

<some:resource> <some:relation> </api/people/john/knows> .
```

In this case, _parent_ resource does not exist as a subject of the relation (i.e. RDF store does not contain any members of this collection), still it is in a relation with some resource and server decided to push that statement.

tpluscode:
> What is a parent here?

In both It has to be subject of the relation. Otherwise I see your concern about ambiguity. "Parent" was poor wording here

tpluscode:
To illustrate my point, consider an entrypoint resource which offers shortcut links to collections I mention in the issue description

```http
GET /api/home

</api/home> hydra:collection </people/>, </events/> .

</people/>
  a hydra:Collection ;
[...]

PR #213, opened 2020-05-10
Relative Template URIs

alien-mcl:
## Summary
Added few more terms to enable different base Uri resolution mechanism of Iri template expansion products which should address issue #208:

- resolveRelativeTo
- BaseUriSource
- Rfc3986Based
- LinkContext

## More details

During a discussion on issue #208 an interesting concept of Iri templates provided on the API documentation level was introduced. While very interesting and tempting, there was no clean way of providing relative Iri templates that could be resolved properly against inlined resources. 
This pull request addresses this issue by introducing some terms that could change the default client behavior while resolving relative Uri that are products or Iri template expansion.

### DIFF
--- a/spec/latest/core/core.jsonld
+++ b/spec/latest/core/core.jsonld
@@ -545,6 +545,36 @@
       "vs:term_status": "testing"
     },
     {
+      "@id": "hydra:resolveRelativeTo",
+      "@type": "rdf:Property",
+      "label": "relative Uri resolution",
+      "domain": "hydra:IriTemplate",
+      "range": "hydra:BaseUriSource",
+      "vs:term_status": "testing"
+    },
+    {
+      "@id": "hydra:BaseUriSource",
+      "@type": "hydra:Class",
+      "subClassOf": "hydra:Resource",
+      "label": "Base Uri source",
+      "comment": "Provides a base abstract for base Uri source for Iri template resolution.",
+      "vs:term_status": "testing"
+    },
+    {
+      "@id": "hydra:Rfc3986Based",
+      "@type": "hydra:BaseUriSource",
+      "label": "RFC 3986 based",
+      "comment": "States that the base Uri should be established using RFC 3986 algorithm.",
+      "vs:term_status": "testing"
+    },
+    {
+      "@id": "hydra:LinkContext",
+      "@type": "hydra:BaseUriSource",
+      "label": "Link context",
+      "comment": "States that the base Uri should be established via a link context.",
+      "vs:term_status": "testing"
+    },
+    {
       "@id": "hydra:offset",
       "@type": "rdf:Property",
       "label": "skip",
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -1040,24 +1040,22 @@
       -->
     </pre>
 
-    <p>Iri expansion should be performed with respect to the specification
-      behind the iri template type (which is RFC 6570 by default), and the
-      product of this process SHOULD be an Iri. In may happen that this
-      product is a relative Uri. In such a case client by default SHOULD
-      stick to RFC 3986. It is due to fact that most RDF serialization
-      allowing other that absolute Iris will behave that way and the result
-      of Iri template expansion should be compliant. Still, it may be
-      preferred to use other base Uri for the expansion process, which
+    <p>IRI expansion should be performed with respect to the specification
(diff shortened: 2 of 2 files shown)


spec/latest/core/index.html
```diff
@@ -1039,6 +1039,44 @@ <h3>Templated Links</h3>
...
+      product is a relative Uri. In such a case client by default SHOULD
+      stick to RFC 3986. It is due to fact that most RDF serialization
+      allowing other that absolute Iris will behave that way and the result
+      of Iri template expansion should be compliant. Still, it may be
+      preferred to use other base Uri for the expansion process, which
+      makes a <i>resolveRelativeTo</i> term useful. It allows to switch the
+      Iri template expansion algorithm so the base Uri is established using
+      current link context, which is a subject of the relation pointing to
+      <i>IriTemplate</i> instance. In case that subject is a relative Uri,
+      default behavior SHOULD be used as fallback.</p>
```
asbjornu, line 1055: I took the liberty to reword the paragraph a bit and made `IRI` and `URI` abbreviations consistenly all-uppercase.

```suggestion
    <p>IRI expansion should be performed with respect to the specification
      behind the IRI template type (RFC 6570 by default), and the product of
      this process SHOULD be an IRI. When the produced IRI is relative, a case
      client SHOULD stick to RFC 3986 to be compatible with most RDF
      serializations that support relative IRIs. Still, it may be preferred to
      use another base URI for the expansion process, which makes a
      <i>resolveRelativeTo</i> term useful. It allows to switch the IRI
      template expansion algorithm so the base URI is established using the
      current link context, which is a subject of the relation pointing to an
      <i>IriTemplate</i> instance. In case that the subject is a relative URI,
      the default behavior SHOULD be used as fallback.
    </p>
```

spec/latest/core/index.html
```diff
@@ -1039,6 +1039,44 @@ <h3>Templated Links</h3>
...
+
+    <p>The example below allows to hook the product of an Iri template
+      expansion to the <i>http://api.example.com/an-issue/</i> resource
+      as it's base Uri, which further enables the <i>some:operation</i> to
+      be moved to i.e. Api documentation level rather to inline it.</p>
```
asbjornu, line 1060: I made a couple of changes to the wording and uppercased abbreviations here as well.

```suggestion
    <p>The example below allows to make the product of an IRI template
      expansion relative to the <i>http://api.example.com/an-issue/</i> resource
      by using it as its base URI, which further enables the
      <i>some:operation</i> to be moved to i.e. API documentation level rather
      to inline it.</p>
```

spec/latest/core/index.html
```diff
@@ -1039,6 +1039,44 @@ <h3>Templated Links</h3>
...
       -->
     </pre>
+
+    <p>Iri expansion should be performed with respect to the specification
+      behind the iri template type (which is RFC 6570 by default), and the
+      product of this process SHOULD be an Iri. In may happen that this
+      product is a relative Uri. In such a case client by default SHOULD
+      stick to RFC 3986. It is due to fact that most RDF serialization
+      allowing other that absolute Iris will behave that way and the result
+      of Iri template expansion should be compliant. Still, it may be
```
tpluscode, line 1049: > It is due to fact that most RDF serialization allowing other that absolute Iris will behave that way and the result of Iri template expansion should be compliant

what did you mean here?

tpluscode: Also, I would specifically mention that sections `5.1.3` and `5.1.4` must be followed to **establish base URI** as the RFC puts it. 

Those are the relevant ones as we discussed that neither `5.1.1` and `5.1.2` can apply in our context.

spec/latest/core/core.jsonld
```diff
@@ -544,6 +544,36 @@
...
+      "@type": "hydra:BaseUriSource",
+      "label": "RFC 3986 based",
+      "comment": "States that the base Uri should be established using RFC 3986 algorithm.",
+      "vs:term_status": "testing"
+    },
+    {
+      "@id": "hydra:LinkContext",
+      "@type": "hydra:BaseUriSource",
+      "label": "Link context",
+      "comment": "States that the base Uri should be established via a link context.",
```
tpluscode, line 574:
```suggestion
      "comment": "States that the link's context IRI, as defined in RFC 5988, should be used as the base Uri.",

```

tpluscode (approved):

Issue #216, opened 2020-05-26
How to document forbidden dereferencability


alien-mcl:
## Documenting forbidden de-referencability

In some cases there is a need to disallow client from de-referencing a resource. Currently there is no way of doing that and `hydra:Resource` implies a resource is (at least virtually) de-referencable (client can i.e. HTTP GET it).

When a resource has a _callable_ IRI (http[s] scheme based one), server may know that it won't support GET operation on this URI, thus it may be required to document that fact.

## Proposed solutions

I currently see 3 viable solutions.

1. Downgrade `rdfs:range` of `hydra:expects` and `hydra:returns` (and possibly `hydra:supportedClass`?) to `rdf:Resource` instead of `hydra:Resource`. According to W3C HTTP specification client cannot assume that HTTP/HTTPS based URIs can be called with GET method, thus API documentation won't encourage client to do so as opposite to currently used `hydra:Resource`.
This may be a breaking change as it may affect existing hydra driven implementations, especially those using RDF reasoning process. This makes it simple to introduce external schema vocabularies like SHACL, OWL and non-rdf based ones.

2. Introduce new term that explicitly forbids client from de-referencing a resource. This won't break anything that is using current hydra version and is somehow consistent with RDF (you need to explicitly say something) and linked data paradigm. It may also open gates for future features like disallowing operations or operation precedence/hiding.

3. Introduce hydra vocabulary profiles or levels, i.e. level 0 and 1 where level 1 would be a default current hydra version and level 0 would be an alternative that does not imply a hydra:Resource is de-referencable (or event RDF based). This could also enable hydra for future profiles/levels using external schema vocabularies like SCHAL, OWL and non-rdf based ones.

tpluscode:
I would definitely have 1️⃣ over 2️⃣ 

---

The third point however I find rather orthogonal. I'm not sure the "profiles" have anything to do with dereferencability of the API descriptions. More about the requirements from the client. I would see this as an open-ended set of "feature-packs" that a server may implement in addition to core. Here are some ideas:

1. Hydra Core
   * Only the base terms to put an API together
2. Standard Profile
   * hydra:Class
2. SHACL profile
   * Shapes for `expects`/`returns`
   * maybe also `supportedClass`
3. JSON Schema profile
   * like SHACL, uses a different vehicle to describe requests and responses
4. IANA Media types profile
   * Uses IANA-registered media types. for example the snippet below might denote an operation where PNG and JPG are allowed to be uploaded

    ```
[...]

PR #217, opened 2020-07-18
How to document forbidden dereferencability

alien-mcl:
## Summary

Please find the modifications (removal of `hydra:Resource`) of the several `rdfs:range`, `rdfs:domain` and `rdfs:subClassOf` definitions that should fix imposed resource dereferencability .

## More details

This should address issue #216 as the core issue was that several places in hydra imposed fact that resources are dereferencable, which in many cases was untrue (not to mention the fact that having i.e. an HTTP based URL doesn't mean it can be freely dereferenced).

There are still a few places where `hydra:Resource` is left:
- both `hydra:Link` and `hydra:TemplatedLink` are still `hydra:Resource`. Removing it from `hydra:Link` could break hydra:apiDocumentation as it actually should be dereferencable. Also links defined by API should connect callable resources.
- the `hydra:operation` has still domain of `hydra:Resource`. Operations should be invokable on a callable resource.
- both `hydra:Collection` and `hydra:PartialCollectionView` are still sub-classes of `hydra:Resource`. These are API artificial wrappers over some resource sets and should be callable.
- all hydra links like `hydra:first`/`hydra:last`/`hydra:next`/`hydra:prev` have still `hydra:Resource` as both their domains and ranges. Otherwise the described _interlinked set of resources_ would cease to be, well, interlinked.
- both `hydra:search` and `hydra:freetextQuery` have still `hydra:Resource` as it's domain. If a client came that far to have that link, it probably means that the resource is already callable.

Feel free to participate in the review as this is a major modification that should improve hydra's applicability to other than RDF areas.

### DIFF
--- a/spec/latest/core/core.jsonld
+++ b/spec/latest/core/core.jsonld
@@ -97,15 +97,15 @@
       "@type": "hydra:Class",
       "subClassOf": "rdfs:Resource",
       "label": "Hydra Resource",
-      "comment": "The class of dereferenceable resources.",
+      "comment": "The class of dereferenceable resources by means a client can at least try to obtain a resource but it still should verify received payloads.",
       "vs:term_status": "testing"
     },
     {
       "@id": "hydra:Class",
       "@type": [ "hydra:Resource", "rdfs:Class" ],
-      "subClassOf": [ "hydra:Resource", "rdfs:Class" ],
+      "subClassOf": [ "rdfs:Class" ],
       "label": "Hydra Class",
-      "comment": "The class of Hydra classes. Hydra classes and their instances are dereferenceable resources.",
+      "comment": "The class of Hydra classes.",
       "vs:term_status": "testing"
     },
     {
@@ -148,7 +148,7 @@
       "label": "supported classes",
       "comment": "A class known to be supported by the Web API",
       "domain": "hydra:ApiDocumentation",
-      "range": "hydra:Class",
+      "range": "rdfs:Class",
       "vs:term_status": "testing"
     },
     {
@@ -165,14 +165,14 @@
       "@type": "hydra:Link",
       "label": "supported properties",
       "comment": "The properties known to be supported by a Hydra class",
-      "domain": "hydra:Class",
+      "domain": "rdfs:Class",
       "range": "hydra:SupportedProperty",
       "vs:term_status": "testing"
     },
     {
       "@id": "hydra:SupportedProperty",
       "@type": "hydra:Class",
-      "subClassOf": "hydra:Resource",
+      "subClassOf": "rdfs:Resource",
       "label": "Supported Property",
       "comment": "A property known to be supported by a Hydra class.",
       "vs:term_status": "testing"
@@ -219,7 +219,7 @@
       "label": "supported operation",
       "comment": "An operation supported by instances of the specific Hydra class or the target of the Hydra link",
       "range": "hydra:Operation",
-      "domainIncludes": ["hydra:Class", "hydra:Link", "hydra:TemplatedLink", "hydra:SupportedProperty"],
+      "domainIncludes": ["rdfs:Class", "hydra:Class", "hydra:Link", "hydra:TemplatedLink", "hydra:SupportedProperty"],
       "vs:term_status": "testing"
     },
     {
@@ -234,7 +234,7 @@
     {
       "@id": "hydra:Operation",
       "@type": "hydra:Class",
-      "subClassOf": "hydra:Resource",
+      "subClassOf": "rdfs:Resource",
       "label": "Operation",
       "comment": "An operation.",
       "vs:term_status": "testing"
@@ -254,8 +254,8 @@
       "label": "expects",
       "comment": "The information expected by the Web API.",
       "domain": "hydra:Operation",
-      "range": "hydra:Resource",
-      "rangeIncludes": ["hydra:Resource", "hydra:Class"],
+      "range": "rdfs:Resource",
+      "rangeIncludes": ["rdfs:Resource", "hydra:Resource", "rdfs:Class", "hydra:Class"],
       "vs:term_status": "testing"
     },
     {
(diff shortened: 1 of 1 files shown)


spec/latest/core/core.jsonld
```diff
@@ -165,14 +165,14 @@
...
-      "domain": "hydra:Class",
+      "domain": "rdfs:Class",
       "range": "hydra:SupportedProperty",
       "vs:term_status": "testing"
     },
     {
       "@id": "hydra:SupportedProperty",
       "@type": "hydra:Class",
-      "subClassOf": "hydra:Resource",
+      "subClassOf": "rdfs:Resource",
```
tpluscode, line 175: I think this is now superfluous and can be removed. `rdfs:Resource` will be the implicit super class

spec/latest/core/core.jsonld
```diff
@@ -234,7 +234,7 @@
     {
       "@id": "hydra:Operation",
       "@type": "hydra:Class",
-      "subClassOf": "hydra:Resource",
+      "subClassOf": "rdfs:Resource",
```
tpluscode, line 237: Ditto, to remove completely

spec/latest/core/core.jsonld
```diff
@@ -340,7 +340,7 @@
       "@type": "hydra:Link",
       "label": "collection",
       "comment": "Collections somehow related to this resource.",
-      "domain": "hydra:Resource",
+      "domain": "rdfs:Resource",
```
tpluscode, line 343: I think this one should stay. If it's an target of a link then it would have to be a dereferencable resource, right?

spec/latest/core/core.jsonld
```diff
@@ -369,16 +369,16 @@
...
     },
     {
       "@id": "hydra:view",
       "@type": "hydra:Link",
       "label": "view",
       "comment": "A specific view of a resource.",
-      "domain": "hydra:Resource",
-      "range": "hydra:Resource",
+      "domain": "rdfs:Resource",
+      "range": "rdfs:Resource",
```
tpluscode, line 381: I think both of these should stay as `hydra:Resource`

spec/latest/core/core.jsonld
```diff
@@ -264,14 +264,14 @@
...
-      "rangeIncludes": ["hydra:Resource", "hydra:Class"],
+      "range": "rdfs:Resource",
+      "rangeIncludes": ["rdfs:Resource", "hydra:Resource", "rdfs:Class", "hydra:Class"],
       "vs:term_status": "testing"
     },
     {
       "@id": "hydra:Status",
       "@type": "hydra:Class",
-      "subClassOf": "hydra:Resource",
+      "subClassOf": "rdfs:Resource",
```
tpluscode, line 274: Ditto, to remove completely

spec/latest/core/core.jsonld
```diff
@@ -369,16 +369,16 @@
       "label": "member",
       "comment": "A member of the collection",
       "domain": "hydra:Collection",
-      "range": "hydra:Resource",
+      "range": "rdfs:Resource",
```
tpluscode, line 372: Ditto, to remove completely

spec/latest/core/core.jsonld
```diff
@@ -463,7 +463,7 @@
     {
       "@id": "hydra:IriTemplate",
       "@type": "hydra:Class",
-      "subClassOf": "hydra:Resource",
+      "subClassOf": "rdfs:Resource",
```
tpluscode, line 466: Ditto, to remove completely

spec/latest/core/core.jsonld
```diff
@@ -499,7 +499,7 @@
     {
       "@id": "hydra:VariableRepresentation",
       "@type": "hydra:Class",
-      "subClassOf": "hydra:Resource",
+      "subClassOf": "rdfs:Resource",
```
tpluscode, line 502: Ditto, to remove completely

spec/latest/core/core.jsonld
```diff
@@ -530,7 +530,7 @@
     {
       "@id": "hydra:IriTemplateMapping",
       "@type": "hydra:Class",
-      "subClassOf": "hydra:Resource",
+      "subClassOf": "rdfs:Resource",
```
tpluscode, line 533: Ditto, to remove completely

spec/latest/core/core.jsonld
alien-mcl: It's a domain. I think a link could be placed between a resource that may be not dereferencable and another resource that is actually dereferencable, i.e. inlined collection member, which may not be dereferencable and still may point to something that can be called.

spec/latest/core/core.jsonld
```diff
@@ -97,15 +97,15 @@
       "@type": "hydra:Class",
       "subClassOf": "rdfs:Resource",
       "label": "Hydra Resource",
-      "comment": "The class of dereferenceable resources.",
+      "comment": "The class of dereferenceable resources by means a client can at least try to obtain a resource but it still should verify received payloads.",
```
asbjornu, line 100: Just a minor tweak to the comment so it reads more easily, imho.

```suggestion
      "comment": "The class of dereferenceable resources a client can attempt to dereference; however, the received responses should still be verified.",
```

spec/latest/core/core.jsonld
```diff
@@ -97,15 +97,15 @@
...
       "vs:term_status": "testing"
     },
     {
       "@id": "hydra:Class",
       "@type": [ "hydra:Resource", "rdfs:Class" ],
-      "subClassOf": [ "hydra:Resource", "rdfs:Class" ],
+      "subClassOf": [ "rdfs:Class" ],
       "label": "Hydra Class",
-      "comment": "The class of Hydra classes. Hydra classes and their instances are dereferenceable resources.",
+      "comment": "The class of Hydra classes.",
```
asbjornu, line 108: Should we explain what "The class of Hydra classes" actually mean? If a `hydra:Class` is no longer dereferenceable, what good does it do?

alien-mcl: Id' leave it as is. Most of the hydra terms are rooted in hydra:Class - I believe it's enough to have it just to have those hydra terms _separated_.

>>If a hydra:Class is no longer dereferenceable,

We indeed removed this requirements as it is virtually impossible to enforce, but still I'd read something that is a hydra:Class SHOULD be dereferencable (I don't see any good reasons against it). Developers still have a possibility of not having those resource dereferencable.

tpluscode (approved):
LGTM

PR #218, opened 2020-11-01
#178 Make hydra:error compatible with RFC 7807

alien-mcl:
## Summary

Changes are covering new standard JSON-LD context that maps RFC-7807 specific JSON properties to Hydra vocabulary.

## More details

This pull-request should resolve issue #718 by introducing two new things:
- standard JSON-LD context that maps RFC-7807 JSON properties to Hydra core vocabulary
- `errorCode` term that is required to make that output of that standard JSON-LD context a `hydra:Error` resource

It may still valid to drop the second change - the only drawback would be the resulting resource would've been a `hydra:Status` which makes the `hydra:Error` itself somehow obsolete (which is not that bad after all).

### DIFF
--- a/spec/latest/core/core.jsonld
+++ b/spec/latest/core/core.jsonld
@@ -31,6 +31,7 @@
     "Status": "hydra:Status",
     "statusCode": "hydra:statusCode",
     "Error": "hydra:Error",
+    "errorCode": "hydra:errorCode",
     "Resource": "hydra:Resource",
     "operation": "hydra:operation",
     "Collection": "hydra:Collection",
@@ -319,6 +320,15 @@
       "subClassOf": "hydra:Status",
       "label": "Error",
       "comment": "A runtime error, used to report information beyond the returned status code.",
+      "vs:term_status": "testing"
+    },
+    {
+      "@id": "hydra:errorCode",
+      "@type": "rdf:Property",
+      "subPropertyOf": "hydra:statusCode",
+      "label": "error code",
+      "comment": "The HTTP status code that is considered an error",
+      "domain": "hydra:Error",
       "vs:term_status": "testing"
     },
     {
--- a/spec/latest/core/core.jsonld
+++ b/spec/latest/core/core.jsonld
@@ -15,7 +15,7 @@
     "description": "hydra:description",
     "entrypoint": { "@id": "hydra:entrypoint", "@type": "@id" },
     "supportedClass": { "@id": "hydra:supportedClass", "@type": "@vocab" },
-    "Class": "hydra:Class",
+    "Class": "hydra:Class","@id": "hydra:Status",
     "supportedProperty": { "@id": "hydra:supportedProperty", "@type": "@id" },
     "SupportedProperty": "hydra:SupportedProperty",
     "property": { "@id": "hydra:property", "@type": "@vocab" },
@@ -283,10 +283,10 @@
       "@id": "hydra:statusCode",
       "@type": "rdf:Property",
       "label": "status code",
-      "comment": "The HTTP status code. It is not recommended to provide this predicate in direct responses.",
+      "comment": "The HTTP status code. It is fully optional and it may happen to be different to actual status code received.",
       "domain": "hydra:Status",
       "range": "xsd:integer",
-      "vs:term_status": "archaic"
+      "vs:term_status": "testing"
     },
     {
       "@id": "hydra:title",
--- a/spec/latest/core/core.jsonld
+++ b/spec/latest/core/core.jsonld
@@ -32,7 +32,6 @@
     "Status": "hydra:Status",
     "statusCode": "hydra:statusCode",
     "Error": "hydra:Error",
-    "errorCode": "hydra:errorCode",
     "Resource": "hydra:Resource",
     "operation": "hydra:operation",
     "Collection": "hydra:Collection",
@@ -284,10 +283,10 @@
       "@id": "hydra:statusCode",
       "@type": "rdf:Property",
       "label": "status code",
-      "comment": "The HTTP status code",
+      "comment": "The HTTP status code. It is not recommended to provide this predicate in direct responses.",
       "domain": "hydra:Status",
       "range": "xsd:integer",
-      "vs:term_status": "testing"
+      "vs:term_status": "archaic"
     },
     {
       "@id": "hydra:title",
@@ -329,15 +328,6 @@
       "subClassOf": "hydra:Status",
       "label": "Error",
       "comment": "A runtime error, used to report information beyond the returned status code.",
-      "vs:term_status": "testing"
-    },
-    {
-      "@id": "hydra:errorCode",
-      "@type": "rdf:Property",
-      "subPropertyOf": "hydra:statusCode",
-      "label": "error code",
-      "comment": "The HTTP status code that is considered an error",
-      "domain": "hydra:Error",
       "vs:term_status": "testing"
     },
     {
--- a/spec/latest/core/core.jsonld
+++ b/spec/latest/core/core.jsonld
@@ -15,7 +15,7 @@
     "description": "hydra:description",
     "entrypoint": { "@id": "hydra:entrypoint", "@type": "@id" },
     "supportedClass": { "@id": "hydra:supportedClass", "@type": "@vocab" },
-    "Class": "hydra:Class","@id": "hydra:Status",
+    "Class": "hydra:Class",
     "supportedProperty": { "@id": "hydra:supportedProperty", "@type": "@id" },
     "SupportedProperty": "hydra:SupportedProperty",
     "property": { "@id": "hydra:property", "@type": "@vocab" },
@@ -283,7 +283,7 @@
       "@id": "hydra:statusCode",
       "@type": "rdf:Property",
       "label": "status code",
-      "comment": "The HTTP status code. It is fully optional and it may happen to be different to actual status code received.",
+      "comment": "The HTTP status code. Please note it may happen this value will be different to actual status code received.",
       "domain": "hydra:Status",
       "range": "xsd:integer",
       "vs:term_status": "testing"
--- a/spec/latest/core/error.jsonld
+++ b/spec/latest/core/error.jsonld
@@ -1,9 +1,11 @@
 {
   "@context": {
     "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
+    "hydra": "http://www.w3.org/ns/hydra/core#",
     "type": "@type",
     "title": "rdfs:label",
     "detail": "rdfs:comment",
+    "status": "hydra:statusCode",
     "instance": { "@id": "rdfs:seeAlso", "@type": "@id" }
   }
 }
--- a/spec/latest/core/error.jsonld
+++ b/spec/latest/core/error.jsonld
@@ -1,11 +1,9 @@
 {
   "@context": {
-    "hydra": "http://www.w3.org/ns/hydra/core#",
     "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
     "type": "@type",
-    "title": "hydra:title",
-    "detail": "hydra:description",
-    "status": "hydra:errorCode",
+    "title": "rdfs:label",
+    "detail": "rdfs:comment",
     "instance": { "@id": "rdfs:seeAlso", "@type": "@id" }
   }
 }
--- a/spec/latest/core/error.jsonld
+++ b/spec/latest/core/error.jsonld
@@ -1,0 +1,11 @@
+{
+  "@context": {
+    "hydra": "http://www.w3.org/ns/hydra/core#",
+    "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
+    "type": "@type",
+    "title": "hydra:title",
+    "detail": "hydra:description",
+    "status": "hydra:errorCode",
+    "instance": { "@id": "rdfs:seeAlso", "@type": "@id" }
+  }
+}
--- a/spec/latest/core/error.jsonld
+++ b/spec/latest/core/error.jsonld
@@ -1,8 +1,9 @@
 {
   "@context": {
+    "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
     "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
     "hydra": "http://www.w3.org/ns/hydra/core#",
-    "type": "@type",
+    "type": { "@id": "rdf:type", "@type": "@id" },
     "title": "rdfs:label",
     "detail": "rdfs:comment",
     "status": "hydra:statusCode",
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -1266,6 +1266,35 @@
       }
       -->
     </pre>
+
+    <p>It is also possible to provide an [[!RFC7807]] compatible</a> error description.
+      To do so, server should provide a <i>application/problem+json</i> response with additional header
+      pointing to <i>http://www.w3.org/ns/hydra/error.context.jsonld</i> context.</p>
+
+    <pre class="example nohighlight" data-transform="updateExample"
+         title="RFC-7807 compatible error description">
+      <!--
+      HTTP/1.1 400 Bad Request
+      Content-Type: application/problem+json
+      Link: <http://www.w3.org/ns/hydra/error.context.jsonld>; rel="http://www.w3.org/ns/json-ld#context"
+
+      {
+        "type": "https://example.com/probs/out-of-credit",
+        "title": "You do not have enough credit.",
+        "detail": "Your current balance is 30, but that costs 50.",
+        "instance": "/account/12345/msgs/abc",
+        "balance": 30,
+        "accounts": ["/account/12345",
+                     "/account/67890"]
+      }
+      -->
+    </pre>
+
+    <p>While this approach makes the response fully compatible with the mentioned specification,
+      properties not defined in the standard Hydra's error context won't be visible for Hydra aware processor.
+      To overcome this, it is possible to declare own context pointed same way, that would combine standard
+      Hydra's error context and a <i>JSON-LD</i>'s with either the <i>@vocab</i> or custom property mappings
+      telling the processor on how to interpret those custom error properties.</p>
   </section>
 
   <section>
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -1284,7 +1284,7 @@
 
     <p>Resources provided may have an additional hint pointing to an <i>Error</i> type like in the example
       above, but it is not mandatory to do so as all resources described with <i>application/problem+json</i>
-      are considered errors.</p>
+      are considered <i>hydra:Error</i>.</p>
 
     <p>It is worth to mention that it may happen (i.e. due to proxy behavior) the value of the <i>status</i> property
       will differ to the one received from the protocol layer.</p>
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -1250,11 +1250,11 @@
       be given an identifier. When dereferenced, the <i>Error</i> resource can provide
       more detailed information or possible ways to resolve the problem, if applicable.<p/>
 
-    <p>Finally, server SHOULD provide error descriptions using an [[!RFC7807]] standard.
-      by using a <i>application/problem+json</i> response. When doing so server also MUST provide
-      an additional header pointing to either built-in Hydra <i>http://www.w3.org/ns/hydra/error</i>
-      context or any JSON-LD context that maps terms <i>type</i>, <i>title</i>, <i>detail</i>
-      and <i>instance</i> same way as the standard one.</p>
+    <p>Finally, the server SHOULD provide error descriptions using an [[!RFC7807]] standard
+      by using an <i>application/problem+json</i> response. When doing so, the server also MUST provide
+      an additional header pointing to either the built-in Hydra <i>http://www.w3.org/ns/hydra/error</i>
+      context or any JSON-LD context that maps the terms <i>type</i>, <i>title</i>, <i>detail</i>
+      and <i>instance</i> the same way as the standard one.</p>
 
     <pre class="example nohighlight" data-transform="updateExample"
          title="RFC-7807 compatible error description">
@@ -1277,8 +1277,8 @@
     </pre>
 
     <p>While the built-in context makes the response fully compatible with the mentioned specification,
-      properties not defined in the standard Hydra's error context won't be visible for Hydra aware processor.
-      To overcome this, it is possible to declare an own context pointed the same way, that would combine standard
+      properties not defined in the standard Hydra's error context won't be visible for Hydra aware processors.
+      To overcome this, it is possible to declare a custom context pointed the same way, that would combine standard
       Hydra's error context and an additional JSON-LD context with either the <i>@vocab</i> or custom property mappings
       telling the processor on how to interpret those custom error properties.</p>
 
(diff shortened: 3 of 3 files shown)


spec/latest/core/core.jsonld
```diff
@@ -321,6 +322,15 @@
...
     },
+    {
+      "@id": "hydra:errorCode",
+      "@type": "rdf:Property",
+      "subPropertyOf": "hydra:statusCode",
+      "label": "error code",
+      "comment": "The HTTP status code that is considered an error",
+      "domain": "hydra:Error",
+      "vs:term_status": "testing"
+    },
```
tpluscode, line 333: Is this necessary? Even looking at the context itself, the property name defined RFC7807 is also simply `status`

spec/latest/core/index.html
```diff
@@ -1266,6 +1266,35 @@ <h3>Description of HTTP Status Codes and Errors</h3>
       }
       -->
     </pre>
+
+    <p>It is also possible to provide an [[!RFC7807]] compatible</a> error description.
+      To do so, server should provide a <i>application/problem+json</i> response with additional header
+      pointing to <i>http://www.w3.org/ns/hydra/error.context.jsonld</i> context.</p>
```
tpluscode, line 1272:
```suggestion
      pointing to <i>http://www.w3.org/ns/hydra/error/context.jsonld</i> context.</p>
```

spec/latest/core/error.jsonld
```diff
@@ -0,0 +1,11 @@
+{
+  "@context": {
+    "hydra": "http://www.w3.org/ns/hydra/core#",
+    "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
+    "type": "@type",
+    "title": "hydra:title",
+    "detail": "hydra:description",
+    "status": "hydra:errorCode",
```
tpluscode, line 8: As stated above, I propose to keep this simple

```suggestion
    "status": "hydra:statusCode",
```

tpluscode (changes requested):

spec/latest/core/core.jsonld
alien-mcl: It was the only approach to have an `hydra:Error` in the result. As I wrote in the description of this pull request - I won't die for it. The only issue I have is that problem+json denotes a, well... problem, an error. The `hydra:Status` does not carry that semantics, `hydra:Error` does.

spec/latest/core/error.jsonld
alien-mcl: Simple is not accurate in this case - let's first decide on the `hydra:Status` and `hydra:Error` issue

spec/latest/core/core.jsonld
tpluscode: Yes, IIRC it was somewhere proposed to actually drop `Error` in favour of `Status`, wasn't it?

spec/latest/core/index.html
serialseb: is this supposed to resolve?

tpluscode: Once it's merged I guess

spec/latest/core/error.jsonld
tpluscode: Which issue is that?

spec/latest/core/core.jsonld
asbjornu: What's the correlation between `hydra:Error` and the response format that will adhere to RFC 7807? I guess it would be interesting for consumers of an API to know whether the response can be parsed according to RFC 7807 or not.

That will of course be possible to indicate with `Content-Type` as well, but knowing which responses to expect from a resource, including which are erroneous and parseable as RFC 7807 may be information that is useful to know before performing an HTTP request.

spec/latest/core/error.jsonld
asbjornu: The one discussed above in https://github.com/HydraCG/Specifications/pull/218/files#r515617548, I assume, @tpluscode?

spec/latest/core/core.jsonld
tpluscode: Check the example in [index.html](https://github.com/HydraCG/Specifications/pull/218/files#diff-91d310ab3038fb1055c658d081cfac162ab60337751b95f239817a8660d0c69cR1277-R1289)

The response is an `application/problem+json`. So yes, parseable according to RFC7807.

If the spec was to state that a Hydra API `MUST` include the error context link in error responses then technically the clients can expect every error response to be a `hydra:Error` shaped as problem+json document.

What do you think about `MUST` @alien-mcl ?

alien-mcl: Currently there is no strong requirement to return either RDF based resource describing a `hydra:Error` or RFC7807 compatible JSON with context. I believe that in case an API is hydra driven, RFC7807 response could have the error context mandatory. I don't know whether we shall recommend RFC7807 over pure RDF response - how do you feel?

spec/latest/core/index.html
alien-mcl: Yes - same way as the core vocabulary context.

spec/latest/core/error.jsonld
alien-mcl: Indeed

spec/latest/core/core.jsonld
alien-mcl: The only thing that supports `hydra:Error` is the fact it matches RFC7807 semantics that is strictly bound to 4XX (and maybe 5XX). The `hydra:Status` resource could be in theory be attached to 3XX (I didn't find the RFC forbidding it - correct me if I'm wrong) or some of the 2XX (i.e. 202 Accepted).
Maybe `hydra:Status` and `hydra:Error` should be disjoined from each other - `hydra:Status` only for API documentation level to describe possible response and `hydra:Error` for 4XX and 5XX inline responses?

tpluscode: > I don't know whether we shall recommend RFC7807 over pure RDF response

I think it makes sense to always return `application/problem+json` with a context link. This way even non-RDF clients or, more importantly, intermediaries will have a better way of understanding error responses.

asbjornu: >I think it makes sense to always return `application/problem+json` with a context link. This way even non-RDF clients or, more importantly, intermediaries will have a better way of understanding error responses.

I agree a `application/problem+json` content type should at least be strongly encouraged, and perhaps even required, for `4xx` and `5xx` responses.

>The `hydra:Status` resource could be in theory attached to 3XX (I didn't find the RFC forbidding it - correct me if I'm wrong) 

Indeed, [`300 Multiple Choices`](https://tools.ietf.org/html/rfc7231#section-6.4.1) responses `SHOULD` contain a payload:

> For request methods other than HEAD, the server SHOULD generate a payload in the 300 response containing a list of representation metadata and URI reference(s) from which the user or user agent can choose the one most preferred.

tpluscode:
Following the discussion between myself, @alien-mcl and @serialseb, here's an example of how an Hydra error response could look like (context inlined)

```
{
  "@context": {
    "hydra": "http://www.w3.org/ns/hydra/core#",
    "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
    "type": "@type",
    "title": "rdfs:label",
    "detail": "rdfs:comment",
    "status": "hydra:statusCode",
    "instance": {
      "@id": "rdfs:seeAlso",
      "@type": "@id"
    }
  },
  "@type": "hydra:Error",
  "type": "https://example.com/probs/out-of-credit",
  "title": "You do not have enough credit.",
  "detail": "Your current balance is 30, but that costs 50.",
  "instance": "/account/12345/msgs/abc",
  "balance": 30,
  "accounts": [
    "/account/12345",
    "/account/67890"
  ]
}
```

1. `rdfs` used in place of hydra terms to make the context more useful to non-Hydra uses
2. `hydra:Error` can be placed in the response body. Mixing `@type` and `type` appears totally fine for JSON-LD parsing

alien-mcl:
I've remapped the context accordingly and made some changes to the spec. In general I've made RFC-8708 errors recommended (SHOULD). This still enables older implementations compatible with the spec.

spec/latest/core/index.html
```diff
@@ -1250,8 +1250,48 @@ <h3>Description of HTTP Status Codes and Errors</h3>
       be given an identifier. When dereferenced, the <i>Error</i> resource can provide
       more detailed information or possible ways to resolve the problem, if applicable.<p/>
 
+    <p>Finally, server SHOULD provide error descriptions using an [[!RFC7807]] standard.
+      by using a <i>application/problem+json</i> response. When doing so server also MUST provide
+      an additional header pointing to either built-in Hydra <i>http://www.w3.org/ns/hydra/error</i>
+      context or any JSON-LD context that maps terms <i>type</i>, <i>title</i>, <i>detail</i>
+      and <i>instance</i> same way as the standard one.</p>
```
asbjornu, line 1257:
```suggestion
    <p>Finally, the server SHOULD provide error descriptions using an [[!RFC7807]] standard
      by using an <i>application/problem+json</i> response. When doing so, the server also MUST provide
      an additional header pointing to either the built-in Hydra <i>http://www.w3.org/ns/hydra/error</i>
      context or any JSON-LD context that maps the terms <i>type</i>, <i>title</i>, <i>detail</i>
      and <i>instance</i> the same way as the standard one.</p>
```

spec/latest/core/index.html
```diff
@@ -1250,8 +1250,48 @@ <h3>Description of HTTP Status Codes and Errors</h3>
...
+        "instance": "/account/12345/msgs/abc",
+        "balance": 30,
+        "accounts": ["/account/12345",
+                     "/account/67890"]
+      }
+      -->
+    </pre>
+
+    <p>While the built-in context makes the response fully compatible with the mentioned specification,
+      properties not defined in the standard Hydra's error context won't be visible for Hydra aware processor.
```
asbjornu, line 1280:
```suggestion
      properties not defined in the standard Hydra's error context won't be visible for Hydra aware processors.
```

spec/latest/core/index.html
```diff
@@ -1250,8 +1250,48 @@ <h3>Description of HTTP Status Codes and Errors</h3>
...
+      To overcome this, it is possible to declare an own context pointed the same way, that would combine standard
```
asbjornu, line 1281:
```suggestion
      To overcome this, it is possible to declare a custom context pointed the same way, that would combine standard
```

spec/latest/core/index.html
```diff
@@ -1250,8 +1250,48 @@ <h3>Description of HTTP Status Codes and Errors</h3>
...
+      Hydra's error context and an additional JSON-LD context with either the <i>@vocab</i> or custom property mappings
+      telling the processor on how to interpret those custom error properties.</p>
+
+    <p>Resources provided may have an additional hint pointing to an <i>Error</i> type like in the example
+      above, but it is not mandatory to do so as all resources described with <i>application/problem+json</i>
+      are considered errors.</p>
+
+    <p>It is worth to mention that the <i>http://www.w3.org/ns/hydra/error</i> context does not map
+      the <i>status</i> property. It is due to possible discrepancies that may occur between actual status returned
+      and the status provided within the payload.</p>
```
asbjornu, line 1291: I don't quite understand the rationale for not mapping `status`?

alien-mcl: The status provided within the payload is not reliable - it may be different than the actual status from the response. I believe the payload should be supplemental for the protocol rather than cause possible discrepancies.

tpluscode: That is irrelevant. The RFC requires that the property and actual status are the same. We should assume that they are 

> Generators MUST use the same status code in the actual HTTP response,
   to assure that generic HTTP software that does not understand this
   format still behaves correctly

asbjornu: Agreed, @tpluscode.

alien-mcl: But this is part of the RFC-7807 spec, not hydra's spec. And I acknowledge this as a flaw - I'm aware that in some circumstances resulting response may still have a different value that the actual status due to i.e. proxy put in the middle that does the change (already saw this in action). I think not mapping the status directly to hydra term is a good move. Actual response should have the `status` property emitted to be compliant with the RFC-7807, but the resulting RDF resource could have that issue fixed.

I see that flaw somehow similar to requirement of having an URL dereferencable - there are circumstances that you cannot guarantee that.

Finally - pragmatic approach would be to see an actual status and not providing that value in the payload by default. This somehow discourages from looking it in the payload.

tpluscode: > Actual response should have the status property emitted to be compliant with the RFC-7807, but the resulting RDF resource could have that issue fixed.
[...]
> Finally - pragmatic approach would be to see an actual status and not providing that value in the payload by default. This somehow discourages from looking it in the payload.

Rightly so; The RFC actually says that the `status` property is optional and only for convenience (or inconvenience?). From [section 3.1](https://tools.ietf.org/html/rfc7807#section-3.1)

> The "status" member, if present, is only advisory

asbjornu: Having `status` in the payload makes it possible to transmit and store the payload otherwise than HTTP without losing any vital information. With the JSON-LD context we define for RFC 7807, we make `application/problem+json` responses available to RDF clients outside the realm of Hydra. I can envision a future where most RDF-enabled clients interpret `application/problem+json` according to Hydra's JSON-LD context. It is therefore vital that we provide a mapping of all properties defined in RFC 7807, regardless of them being optional or not.

tpluscode: Totally. The RFC defines five standard fields. We must provide mapping for each one

alien-mcl: >>Having status in the payload makes it possible to transmit and store the payload otherwise than HTTP without losing any vital information.

Not really - the `type` is vital enough and carries all the details that client might need.

Anyway, I may agree that hydra could be the first one to address an issue of moving from `application/problem+json` to RDF world. We don't have too much options here:
- ignore `status` and leave it unmapped
- map `status` to `hydra:statusCode` and mark it as HTTP only and make it optional as it is in RFC-7807
- map `status` to other _low_level_ vocabulary that we could incorporate here; the only guess I have here would be to use `http:statusCodeNumber` from the HTTP Vocabulary in RDF 1.0, but I didn't see anyone using it; from RDF point of view it would also make the resource an `http:StatusCode` instance; I didn't find anything fitting - maybe `skos:notation` ?

tpluscode: The `type` and every other field in problem+json are in fact optional.

There are 5 fields. We must map all of them. To keep `http` vocabulary as had been decided on other occasions, I vote for using `hydra:statusCode`

In the future, should a specific set of terms arise for Problem Details, we can make our term a subProperty as we did with member assertions

asbjornu: Agreed, @tpluscode.

alien-mcl: I've reintroduced `hydra:statusCode` mapping for the `status` property and made few changes in the spec to note the possible issue with it. I hope it resolves the PR.

spec/latest/core/index.html
```diff
@@ -1250,8 +1250,48 @@ <h3>Description of HTTP Status Codes and Errors</h3>
       be given an identifier. When dereferenced, the <i>Error</i> resource can provide
       more detailed information or possible ways to resolve the problem, if applicable.<p/>
 
+    <p>Finally, the server SHOULD provide error descriptions using an [[!RFC7807]] standard
+      by using an <i>application/problem+json</i> response. When doing so, the server also MUST provide
+      an additional header pointing to either the built-in Hydra <i>http://www.w3.org/ns/hydra/error</i>
+      context or any JSON-LD context that maps the terms <i>type</i>, <i>title</i>, <i>detail</i>
+      and <i>instance</i> the same way as the standard one.</p>
```
asbjornu, line 1257:
```suggestion
      context or any JSON-LD context that maps the terms <i>type</i>, <i>title</i>, <i>detail</i>,
      <i>status</i> and <i>instance</i> the same way as the standard one.</p>
```

spec/latest/core/index.html
```diff
@@ -1250,8 +1250,48 @@ <h3>Description of HTTP Status Codes and Errors</h3>
...
+        "accounts": ["/account/12345",
+                     "/account/67890"]
+      }
+      -->
+    </pre>
+
+    <p>While the built-in context makes the response fully compatible with the mentioned specification,
+      properties not defined in the standard Hydra's error context won't be visible for Hydra aware processors.
+      To overcome this, it is possible to declare a custom context pointed the same way, that would combine standard
+      Hydra's error context and an additional JSON-LD context with either the <i>@vocab</i> or custom property mappings
```
asbjornu, line 1282:
```suggestion
      To overcome this, it is possible to declare a custom context pointed the same way, that would combine Hydra's
      standard error context and an additional JSON-LD context with either the <i>@vocab</i> or custom property mappings
```

spec/latest/core/index.html
```diff
@@ -1250,8 +1250,48 @@ <h3>Description of HTTP Status Codes and Errors</h3>
...
+      telling the processor on how to interpret those custom error properties.</p>
+
+    <p>Resources provided may have an additional hint pointing to an <i>Error</i> type like in the example
+      above, but it is not mandatory to do so as all resources described with <i>application/problem+json</i>
+      are considered errors.</p>
+
+    <p>It is worth to mention that the <i>status</i> property is optional and while it should be the same as the
```
asbjornu, line 1289: All properties are optional in RFC 7807. The guidance around `status` should be given in RFC 7807 and not by Hydra, imho. We are just providing a mapping, not implementation guidance for the RFC 7807 specification.

asbjornu (changes requested):

spec/latest/core/core.jsonld
```diff
@@ -283,7 +283,7 @@
       "@id": "hydra:statusCode",
       "@type": "rdf:Property",
       "label": "status code",
-      "comment": "The HTTP status code",
+      "comment": "The HTTP status code. It is fully optional and it may happen to be different to actual status code received.",
```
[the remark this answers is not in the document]
tpluscode: In principle any RDF property is optional unless stated otherwise. I would remove that part

spec/latest/core/index.html
alien-mcl: Latest changes are using `http://www.w3.org/ns/hydra/error` which should resolve with content negotiation  - is that acceptable for you @tpluscode ?

spec/latest/core/core.jsonld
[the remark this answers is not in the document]
alien-mcl: I've reworded this sentence so the user is warned of the consequences. My aim is to somehow discourage to use it when possible.

asbjornu (approved):

spec/latest/core/index.html
```diff
@@ -1250,8 +1250,47 @@ <h3>Description of HTTP Status Codes and Errors</h3>
...
+
+    <pre class="example nohighlight" data-transform="updateExample"
+         title="RFC-7807 compatible error description">
+      <!--
+      HTTP/1.1 400 Bad Request
+      Content-Type: application/problem+json
+      Link: <http://www.w3.org/ns/hydra/error>; rel="http://www.w3.org/ns/json-ld#context"
+
+      {
+        "type": "https://example.com/probs/out-of-credit",
```
lanthaler, line 1267: Given that `type` is mapped to `@type` in the error context this will cause an error

tpluscode: Works just fine in the playground. Also in the legacy 1.0 version

alien-mcl: Actually - in 1.0 it seems to throw, see https://json-ld.org/playground/1.0/#startTab=tab-expanded&json-ld=%7B%22%40context%22%3A%7B%22rdfs%22%3A%22http%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23%22%2C%22hydra%22%3A%22http%3A%2F%2Fwww.w3.org%2Fns%2Fhydra%2Fcore%23%22%2C%22type%22%3A%22%40type%22%2C%22title%22%3A%22rdfs%3Alabel%22%2C%22detail%22%3A%22rdfs%3Acomment%22%2C%22status%22%3A%22hydra%3AstatusCode%22%2C%22instance%22%3A%7B%22%40id%22%3A%22rdfs%3AseeAlso%22%2C%22%40type%22%3A%22%40id%22%7D%7D%2C%22type%22%3A%22https%3A%2F%2Fexample.com%2Fprobs%2Fout-of-credit%22%2C%22%40type%22%3A%22http%3A%2F%2Fwww.w3.org%2Fns%2Fhydra%2Fcore%23Error%22%2C%22title%22%3A%22You%20do%20not%20have%20enough%20credit.%22%2C%22detail%22%3A%22Your%20current%20balance%20is%2030%2C%20but%20that%20costs%2050.%22%2C%22instance%22%3A%22%2Faccount%2F12345%2Fmsgs%2Fabc%22%2C%22balance%22%3A30%2C%22accounts%22%3A%5B%22%2Faccount%2F12345%22%2C%22%2Faccount%2F67890%22%5D%7D. Thank's @lanthaler - I'll need to make some more experiments with it

tpluscode: That's weird. I tried this earlier and I thought it was working 😔. 

We can recommend JSON-LD 1.1 but probably a bad idea to have it as a requirement?

Alternatively `"type": ["https://example.com/probs/out-of-credit", "hydra:Error"]` is fine in JSON-LD terms but I thank is invalid problem+json

alien-mcl: Good question - can a problem+json point to multiple types using array ?

asbjornu: Nope, `type` is a `string` that should be in the form of a URI. Is it impossible to map this in JSON-LD somehow? Or would that require Hydra to demand JSON-LD version 1.1?

alien-mcl: What is the state of JSON-LD 1.0 specification? When I browsed that spec all I can see a popup with "This version is outdated!". Does it mean 1.0 is deprecated and 1.1 should be used? Like in case of RFC 7230 that makes the older one, RFC 2616 obsolete?

lanthaler: I would suggest to not require Hydra clients to understand JSON-LD 1.1 just for this. The pragmatic solution would be to map RFC7807's `type` to something else than `@type` aka `rdf:type`. Perhaps `hydra:errorType`... or introduce a separate vocab for RFC 7807 altogether

tpluscode: Obviously! Thank you @lanthaler

spec/latest/core/error.jsonld
```diff
@@ -0,0 +1,11 @@
+{
+  "@context": {
+    "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
+    "hydra": "http://www.w3.org/ns/hydra/core#",
+    "type": "@type",
```
tpluscode, line 5:
```suggestion
    "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
    "type": { "@id": "rdf:type", "@type": "@id" },
```

As suggested by @lanthaler, this makes this context usable by JSON-LD 1.0 processors

tpluscode, line 5:
```suggestion
    "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
    "type": { "@id": "rdf:type", "@type": "@id" },
```

As suggested by @lanthaler, this makes this context usable by JSON-LD 1.0 processors

tpluscode (changes requested):

spec/latest/core/index.html
alien-mcl: I don't see how having a separate vocab solves the type mapping issue - there would also be some term to be mapped and the context to be usage agnostic.
Also mapping it to `rdf:type` feels like a hack.
Enforcing JSON-LD 1.1 is neither an option for me.
I'd go with a statement that response of type `application/problem+json` using this context should be considered `hydra:Error` and just make a note for the example that the `"@type": "http://www.w3.org/ns/hydra/core#Error"` is optional and JSON-LD 1.1 compatible. Or maybe we can get rid of it completely for sake of simplicity.

tpluscode: > for sake of simplicity.

`rdf:type` **is** the simple solution. It just works and requires no sacrifice or weird conditionals in the spec. Why do you say it's hack? `@type` is just a syntactic sugar equivalent to using `rdf:type` directly.

asbjornu: `rdf:type` seems to be just the type of glue we need here. 👍🏼 from me.

alien-mcl: OK, `rdf:type` it is - I'll update the PR ASAP

tpluscode (approved):

Issue #227, opened 2021-04-07
Supported property values source


alien-mcl:
## Describe the requirement

There are circumstances in which API would like to announce source of values for a specific supported property.
While Hydra is not focusing on data structure descriptions, binding between properties and operations feeding these properties feels like in the area of interest for Hydra Core Vocabulary.

## Hydra-agnostic example

Imagine a UI that shows an auto-generated form based on what API says, i.e. an issue. Between simple literal properties like title and description there is also an assignee. In order to avoid manual typing of person's details I'd like to be able to pick one from a list. Now the tricky part - how to drive that list (and the fact there is that list available).

## Proposed solutions

My idea is to extend a bit a specification by inserting a new section in paragraph 5. Advanced Concepts. That bit of a description would describe _special_ behavior that client would need to incorporate in case of either _hydra:collection_ or _hydra:search_ predicates encountered either on instances of _hydra:SupportedProperty_ or _rdf:Property_. The former would let the client to discover that the collection provided feeds values for that very property and the latter provides a type-ahead-like list of values for that very property.

```turtle
<> a hydra:ApiDocumentation;
  hydra:supportedClass api:Issue.
api:Issue a rdf:Class;
  hydra:supportedProperty [
    hydra:property api:assignee;
    hydra:search [
      hydra:template "http://temp.uri/api/user{?search}";
      hydra:variable "search";
      hydra:property hydra:freetextQuery;
      hydra:required "true"^^xsd:boolean
    ]
  ].
```

These details could be provided either on API Documentation level and inlined mandating the client to use the most specific source possible. This would mean that if an inlined declaration is available in the request that drove the form creation, that declaration should be used. Otherwise API documentation level declaration can be used as fallback.

PR #229, opened 2021-04-11
#227 Supported property values source

alien-mcl:
## Summary

Added a paragraph describing on how collection or search can be used to point to supported property's values source.

## More details

This PR addresses issue #227 and describes on how already existing terms can be used to achieve what's needed. No additional terms are defined.


tpluscode:
It would be nice to have an example snippet here

tpluscode (approved):

PR #230, opened 2021-04-12
replaced possibleStatus by statusCode in example

cKlee:
## Summary

Replace property possibleStatus by statusCode in example 27.

## More details

Property `possibleStatus` has range `hydra:Status`. Example shows an integer value. Matching property must be `statusCode`.


cKlee:
Why is this closed and not merged?

alien-mcl:
Oh sorry - I didn't hit the merge button!

Issue #233, opened 2021-04-13
Strongly typed collections in API documentation


alien-mcl:
## Describe the requirement

I was wondering on how to describe a strongly typed (closed generic) collection in API documentation.

## Hydra-agnostic example

There are situations when client could benefit from known in advance what kind of collection members are in the collection.
One of the reason would be answer to question "what should I do to obtain a list of users".

## Proposed solutions

There are two approaches. One would be to include a piece of a collection mentioned by the API documentation with that documentation itself:
```turtle
  @base <http://temp.uri/api> .
  </api> a hydra:ApiDocumentation;
    hydra:supportedClass api:User.
  hydra:collection </api/users>.
  </api/users> hydra:memberAssertion [
    hydra:predicate rdf:type;
    hydra:object api:User].
```

This does not require any new terms, but it may be unwanted to mix metadata (an API documentation) with pieces of actual resources. Alternative would be a new term that would allow to more direct and abstract description:

```turtle
  @base <http://temp.uri/api> .
  </api> a hydra:ApiDocumentation;
    hydra:supportedClass api:User;
    api:users </api/users> .
  api:users a hydra:Link;
    hydra:supportedOperation [
      hydra:method "GET";
      hydra:returns [
        rdfs:subClassOf hydra:Collection;
        hydra:memberAssertionEquivalent [
          hydra:predicate rdf:type;
          hydra:object api:User]
      ]
  ] .
```

The example above introduces a hypothetical `hydra:memberAssertionEquivalent` term (name is a subject to change).
Another alternative would be:

```turtle
  @base <http://temp.uri/api> .
  </api> a hydra:ApiDocumentation;
    hydra:supportedClass api:User;
    api:users </api/users> .
  api:users a hydra:Link;
    hydra:supportedOperation [
      hydra:method "GET";
      hydra:returns hydra:Collection;
      hydra:memberAssertionEquivalent [
          hydra:predicate rdf:type;
          hydra:object api:User]
      ]
  ] .
```

A brand new term is introduced and attached in a different place (less nesting) compared to the previous example. Slight modification for the two above would require reusing existing `hydra:memberAssertion` with it's domain modified (currently `hydra:Collection`). Both of the cases above are abstract descriptions and there are no statements in which the actual resource instance is a subject for.

Feel free to deliberate on the best approach.

tpluscode:
I like what you write in the last paragraph.

If the domain was changed to loosely allow both `Class` and `Collection`

```diff
[...]

PR #240, opened 2021-06-05
Added support for strongly typed collections

alien-mcl:
## Summary

This pull request add some changes to the spec that should enable hydra powered APIs to describe strongly typed collections.

## More details

There is an issue #233 that should be addressed by this pull request. While the changes are somehow small in numbers, there is another possibly a breaking change as one of the terms has it's domain changed.

### DIFF
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -1032,7 +1032,7 @@
 
     <pre class="example nohighlight"
          data-transform="updateExample"
-         title="Member assertion block describes relation with another resource">
+         title="Member assertion describes relation with another resource">
       <!--
       {
         "@context": "http://www.w3.org/ns/hydra/context.jsonld",
@@ -1071,14 +1071,19 @@
       <a href="https://www.w3.org/TR/rdf-schema/"><code>rdf</code></a> terms.</p>
 
     <p>It is also possible to use <i>memberAssertion</i> predicate on the API documentation level,
-      by attaching this predicate to subclasses of the <i>Collection</i>, like in the example below:</p>
+      by attaching this predicate to subclasses of the <i>Collection</i>, like in the example below.
+      Clients would understand that all members of collections which are instances of <i>api:UserCollections</i>
+      would in fact have <i>rdf:type api:User</i>.</p>
 
     <pre class="example nohighlight"
          data-transform="updateExample"
          title="Member assertion block describes strongly typed collection">
       <!--
       {
-        "@context": "http://www.w3.org/ns/hydra/context.jsonld",
+        "@context": [
+          "http://www.w3.org/ns/hydra/context.jsonld",
+          { "api": "http://api.example.com/api/documentation#" }
+        ],
         "@id": "http://api.example.com/api/documentation",
         "@type": "ApiDocumentation",
         "supportedClass": [api:User, api:UserCollection],
@@ -1086,12 +1091,24 @@
           "subClassOf": "Collection",
           ****"memberAssertion": {
             "property": "rdf:type",
-            "object": "http://api.example.com/api/documentation#User"
+            "object": "api:User"
           }****,
         }
       }
       -->
     </pre>
+
+    <p>It is worth to mention that a strongly typed collection instance can have its own
+      member assertions. n such scenario, both <i>API documentation</i> level and instance
+      level assertions should be combined neither makes the other obsolete.</p>
+
+    <p>More complex scenario would involve a class hierarchy, in which each class can
+      carry additional member assertions compared to it's base class. To discover all
+      available member assertion blocks client SHOULD traverse whole class hierarchy
+      to gather all of the member assertions. In order to take such a burden from clients
+      it is strongly recommended to provide all member assertion blocks to be provided
+      on each class level including it's base class blocks so the client does not have
+      to perform this traversing behavior.</p>
   </section>
 
   <section>
--- a/spec/latest/core/index.html
+++ b/spec/latest/core/index.html
@@ -1032,7 +1032,7 @@
 
     <pre class="example nohighlight"
          data-transform="updateExample"
-         title="Manages block describes relation with another resource">
+         title="Member assertion block describes relation with another resource">
       <!--
       {
         "@context": "http://www.w3.org/ns/hydra/context.jsonld",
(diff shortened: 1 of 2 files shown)


spec/latest/core/index.html
```diff
@@ -1032,7 +1032,7 @@ <h4>Member assertions</h4>
 
     <pre class="example nohighlight"
          data-transform="updateExample"
-         title="Manages block describes relation with another resource">
+         title="Member assertion block describes relation with another resource">
```
tpluscode, line 1035:
```suggestion
         title="Member assertion describes relation with another resource">
```

spec/latest/core/index.html
```diff
@@ -1069,6 +1069,29 @@ <h4>Member assertions</h4>
     <p class="note">It's important to point out that the <code>subject</code>, <code>property</code>
       and <code>object</code> predicates are defined within the Hydra namespace and are not
       <a href="https://www.w3.org/TR/rdf-schema/"><code>rdf</code></a> terms.</p>
+
+    <p>It is also possible to use <i>memberAssertion</i> predicate on the API documentation level,
+      by attaching this predicate to subclasses of the <i>Collection</i>, like in the example below:</p>
```
tpluscode, line 1074:
```suggestion
      by attaching this predicate to subclasses of the <i>Collection</i>, like in the example below.
      Clients would understand that all members of collections which are instances of <i>api:UserCollection</i>`
      would in fact have <i>rdf:type api:User</i></p>
```

spec/latest/core/index.html
```diff
@@ -1069,6 +1069,29 @@ <h4>Member assertions</h4>
...
+      {
+        "@context": "http://www.w3.org/ns/hydra/context.jsonld",
+        "@id": "http://api.example.com/api/documentation",
+        "@type": "ApiDocumentation",
+        "supportedClass": [api:User, api:UserCollection],
+        "api:UserCollection": {
+          "subClassOf": "Collection",
+          ****"memberAssertion": {
+            "property": "rdf:type",
+            "object": "http://api.example.com/api/documentation#User"
```
tpluscode, line 1089:
```suggestion
            "object": "api:User"
```

spec/latest/core/index.html
```diff
@@ -1069,6 +1069,29 @@ <h4>Member assertions</h4>
...
+
+    <p>It is also possible to use <i>memberAssertion</i> predicate on the API documentation level,
+      by attaching this predicate to subclasses of the <i>Collection</i>, like in the example below:</p>
+
+    <pre class="example nohighlight"
+         data-transform="updateExample"
+         title="Member assertion block describes strongly typed collection">
+      <!--
+      {
+        "@context": "http://www.w3.org/ns/hydra/context.jsonld",
```
tpluscode, line 1081: Adding this so that we can have a valid `api:` prefix as used below

```suggestion
        "@context": [
          "http://www.w3.org/ns/hydra/context.jsonld",
          {
            "api": "http://api.example.com/api/documentation#"
          }
        ],
```

tpluscode (changes requested):
Awesome, like it overall. Just suggested some editiorial changes

tpluscode:
Something just occurred to me. I think it we should add a paragraph describing the case when an instance of a strongly type can have its own member assertions. In that scenario all those assertions should be combined, right?

tpluscode:
Ah, and one more, similar case. We should clarify the behaviour when there are multiple (sub)classes which may have member assertions. I'd say that clients and server should walk up the `rdfs:subClassOf` hierarchy and combine them all. 

To sum up with an example

```turtle
:ArticleCollection
  rdfs:subClassOf hydra:Collection ;
  hydra:memberAssertion [
    hydra:property rdf:type ;
    hydra:object :Article ;
  ] ;
.

:DraftArticleCollection
  rdfs:subClassOf :ArticleCollection ;
  hydra:memberAssertion [
    hydra:property :status ;
    hydra:object :Draft ;
  ] ;
.

:KarolsDraftsCollection
  rdfs:subClassOf :DraftArticleCollection ;
  hydra:memberAssertion [
    hydra:property acl:owner ;
    hydra:object </person/Karol> ;
  ] ;
.

</person/Karol/drafts> a :KarolsDraftsCollection .
```

By combining all the member assertions above we could generate a query like below to find all members of the resource `</person/Karol/drafts>`.


```sparql
SELECT ?member {
  </person/Karol/drafts> hydra:member ?member .

  # asserted by :ArticleCollection 
  ?member a :Article .

  # asserted by :DraftArticleCollection
  ?member :status :Draft .

  # asserted by :KarolsDraftsCollection
  ?member acl:owner </person/Karol> .
}

**Comment by alien-mcl** [comment]
I've added both paragraphs, but I think clients should not be overweighed with such a complex tasks as class hierarchy traversing.

**Comment by alien-mcl** [comment]
@tpluscode could you please review changes you've requested? I've got another bunch of changes made on this branch I'd like to proceed with (brand new diagram and made the terms stable rather than testing)

File: spec/latest/core/index.html [hunk 5]
```diff
@@ -1069,6 +1069,46 @@ <h4>Member assertions</h4>
...
+          ****"memberAssertion": {
+            "property": "rdf:type",
+            "object": "api:User"
+          }****,
+        }
+      }
+      -->
+    </pre>
+
+    <p>It is worth to mention that a strongly typed collection instance can have its own
```
tpluscode, line 1101:
```suggestion
    <p>It is worth mentioning that a strongly typed collection instance can have its own
```

spec/latest/core/index.html
```diff
@@ -1069,6 +1069,46 @@ <h4>Member assertions</h4>
...
+      member assertions. In such scenario, both <i>API documentation</i> level and instance
+      level assertions should be combined as neither makes the other obsolete.</p>
+
+    <p>More complex scenario would involve a class hierarchy, in which each class can
+      carry additional member assertions compared to it's base class. To discover all
+      available member assertion blocks client SHOULD traverse whole class hierarchy
+      to gather all of the member assertions. In order to take such a burden from clients
+      it is strongly recommended to provide all member assertion blocks to be provided
+      on each class level including it's base class blocks so the client does not have
+      to perform this traversing behavior.</p>
```
tpluscode, line 1111: I don't think this makes the much sense. Even if an API duplicates all assertions on all levels of a class hierarchy, the client does not have a way of knowing that. Thus, they will have to traverse the hierarchy nonetheless.

I think I would change this paragraph and require only that a client inspects explicit `rdf:type` of a collection resource. In the example below the client would only discover the `hydra:manages` of `PersonCollection`

```turtle
# resource
</people> a api:PersonCollection, hydra:Collection .

# api docs
api:PersonCollection
  rdfs:subClassOf api:AgentCollection ;
  hydra:manages [ hydra:property rdf:type ; hydra:object foaf:Person ] ;
.

api:AgentCollection
  rdfs:subClassOf hydra:Collection ;
  hydra:manages [ hydra:property rdf:type ; hydra:object foaf:Agent ] ;
.
```

Technically, a client could perform RDFS reasoning but in practice I would suggest that the server did that and served all materialised RDF types instead, to relieve the clients from performing additional processing

tpluscode (changes requested):

spec/latest/core/index.html
alien-mcl: Hmm - I see your point. I've softened this required from _SHOULD_ to _MAY_. Strong recommendation for the server to provide all the details is still in force, but I do not see a way of forcing developers to do so. The client still can do the traversing though.

daryazzarfa1988 (approved):

PR #242, opened 2021-07-13
Fix JSON-LD syntax in examples

jaw111:
Fix syntax errors in JSON-LD examples


alien-mcl (approved):

alien-mcl:
Thanks for corrections!

